PR #107, opened 2021-03-11
Prepare to onboard to prow

jlpadilla:
**Related Issue:**  open-cluster-management/backlog#9582

### Description of changes
- Add `Makefile.prow`
- Add `Dockerfile.prow`
- Update Dockerfile to be consistent with Dockerfile.prow

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,10 @@
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.2
 
+WORKDIR /go/src/github.com/open-cluster-management/search-collector
+COPY . .
+RUN CGO_ENABLED=0 GOGC=25 go build -trimpath -o main main.go
+
+
 ARG VCS_REF
 ARG VCS_URL
 ARG IMAGE_NAME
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,5 +1,10 @@
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.2
 
+WORKDIR /go/src/github.com/open-cluster-management/search-collector
+COPY . .
+RUN CGO_ENABLED=0 GOGC=25 go build -trimpath -o main main.go
```
mqhackett, line 5: For reference, these are the resource requests set for each step of the build. We can modify this by changing our openshift release repo if necessary. The recommend/default are:
```
resources:
  '*':
    requests:
      cpu: 100m
      memory: 200Mi
```

DCO
```diff
@@ -0,0 +1,37 @@
+Developer Certificate of Origin
```
mqhackett, line 1: This is new. Needed for all projects?

mqhackett:
I want to merge this PR so that prow has it's `.prow` files to use for the new pipeline. We can continue making small PRs to Makefile and Dockerfile as necessary.

mqhackett (approved):

PR #114, opened 2021-03-17
Index Argo applications using structs

KevinFCormier:
**Related Issue:**  open-cluster-management/backlog#9514

### Description of changes
Indexes Argo applications, defining structs that include the fields we are interested in
Adds a test case for indexing of Argo applications


pkg/transforms/argoapplication.go
```diff
@@ -0,0 +1,81 @@
...
+	// Destination properties
+	if a.Spec.Destination.Name != "" {
+		node.Properties["destinationName"] = a.Spec.Destination.Name
+	}
+	if a.Spec.Destination.Namespace != "" {
+		node.Properties["destinationNamespace"] = a.Spec.Destination.Namespace
+	}
+	if a.Spec.Destination.Server != "" {
+		node.Properties["destinationServer"] = a.Spec.Destination.Server
+	}
```
jlpadilla, line 53: This pattern may cause problems if any of these properties is removed during an update.  Let's send the empty value, this guarantees that the existing value is removed.

_Also sending the empty value keeps the data schema always the same, which might help us in the future when we replace Redisgraph._

jlpadilla (approved):

jlpadilla (approved):
Thanks for all the updates!

PR #116, opened 2021-03-22
remove deprecated selfLink

SherinV:
**Related Issue:**  open-cluster-management/backlog#10696

### Description of changes
- Removes selfLink.  It was removed in K8s 1.20

### DIFF
--- a/pkg/reconciler/reconciler_test.go
+++ b/pkg/reconciler/reconciler_test.go
@@ -69,7 +69,6 @@ func createNodeEvents() []tr.NodeEvent {
 	p.Name = "testpod"
 	p.Kind = "Pod"
 	p.Namespace = "default"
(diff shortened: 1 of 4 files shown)


SherinV:
Do we have to remove the ones in the test-data jsons? https://github.com/open-cluster-management/search-collector/blob/0f4bd03f1ae4e331be89063f753e83826d5c4232/test-data/appdeployable.json#L24

jlpadilla (changes requested):
Changes are good, but first we need to remove any usage of the `selfLink` field from `search-api`.

SherinV:
> Changes are good, but first we need to remove any usage of the `selfLink` field from `search-api`.

Yes, facing some issues with committing changes there. The pre-commit hook is failing. Will check that

jlpadilla (approved):

PR #120, opened 2021-04-07
 migrated to v1alpha2

xrajesh:
**Related Issue:**  open-cluster-management/backlog#<ISSUE_NUMBER>

### DIFF
--- a/pkg/transforms/policyreport.go
+++ b/pkg/transforms/policyreport.go
@@ -5,35 +5,36 @@ package transforms
 
 import (
 	"strings"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // PolicyReport report
 type PolicyReport struct {
-	metav1.TypeMeta `json:",inline"`
+	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
-	Results []ReportResults `json:"results"`
+	Results           []ReportResults `json:"results"`
 }
 
 // ReportResults rule violation results
 type ReportResults struct {
-	Policy string `json:"policy"`
-	Message string `json:"message"`
-	Category string `json:"category"`
-	Status string `json:"status"`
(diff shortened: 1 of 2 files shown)


pkg/transforms/policyreport.go
```diff
@@ -5,35 +5,36 @@ package transforms
...
-	Data ReportData `json:"data"`
+	Policy      string     `json:"policy"`
+	Description string     `json:"message"`
+	Category    string     `json:"category"`
+	Result      string     `json:"result"`
+	Properties  ReportData `json:"properties"`
 }
 
 // ReportData rule violation data
 type ReportData struct {
```
zlayne, line 29: Maybe rename to ReportProperties?

pkg/transforms/policyreport.go
```diff
@@ -5,35 +5,36 @@ package transforms
...
-	Policy string `json:"policy"`
-	Message string `json:"message"`
-	Category string `json:"category"`
-	Status string `json:"status"`
-	Data ReportData `json:"data"`
+	Policy      string     `json:"policy"`
+	Description string     `json:"message"`
+	Category    string     `json:"category"`
+	Result      string     `json:"result"`
+	Properties  ReportData `json:"properties"`
```
zlayne, line 25: Should we add timestamp here as well?

pkg/transforms/policyreport.go
```diff
@@ -48,9 +49,9 @@ func PolicyReportResourceBuilder(pr *PolicyReport) *PolicyReportResource {
 	node.Properties["apigroup"] = gvk.Group
 
 	// Extract the properties specific to this type
-	node.Properties["message"] = string(pr.Results[0].Message)
+	node.Properties["message"] = string(pr.Results[0].Description)
```
zlayne, line 52: Would it be less confusing to use description as the property here?

zlayne (approved):

pkg/transforms/policyreport.go
xrajesh: Not sure we want Timestamp in search , until we need it we dont want to cache in redis .

PR #126, opened 2021-04-19
loop through and get all policies in the report

SherinV:
**Related Issue:**  open-cluster-management/backlog#11701

### Description of changes
Make changes to consolidate all policies in one report

### DIFF
--- a/pkg/transforms/policyreport.go
+++ b/pkg/transforms/policyreport.go
@@ -48,8 +48,6 @@ func PolicyReportResourceBuilder(pr *PolicyReport) *PolicyReportResource {
 	node.Properties["apiversion"] = gvk.Version
 	node.Properties["apigroup"] = gvk.Group
 
-	// Overwrite the name to be the namespace - which is same as the cluster the policy affects
-	node.Properties["name"] = pr.Namespace
 	// Total number of policies in the report
 	node.Properties["numInsightPolicies"] = len(pr.Results)
 	// Extract the properties specific to this type
--- a/pkg/transforms/policyreport.go
+++ b/pkg/transforms/policyreport.go
@@ -18,11 +18,11 @@ type PolicyReport struct {
 
 // ReportResults rule violation results
 type ReportResults struct {
-	Policy      string           `json:"policy"`
-	Message     string           `json:"message"`
-	Category    string           `json:"category"`
-	Result      string           `json:"result"`
-	Properties  ReportProperties `json:"properties"`
+	Policy     string           `json:"policy"`
+	Message    string           `json:"message"`
+	Category   string           `json:"category"`
+	Result     string           `json:"result"`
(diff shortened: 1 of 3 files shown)


pkg/transforms/policyreport.go
```diff
@@ -48,12 +48,25 @@ func PolicyReportResourceBuilder(pr *PolicyReport) *PolicyReportResource {
 	node.Properties["apiversion"] = gvk.Version
 	node.Properties["apigroup"] = gvk.Group
 
+	// Overwrite the name to be the namespace - which is same as the cluster the policy affects
+	node.Properties["name"] = pr.Namespace
```
jlpadilla, line 52: If we overwrite the name here, the UI will be broken when displaying the yaml. It will also break the delete action.

FYI @zlayne

zlayne: Yea lets just use the pr.Name (it will be the same as the namespace - we are using the cluster name for both). I am making the necessary changes in insights-client to change the name of policyreports to the cluster name/namespace.

pkg/transforms/policyreport.go
```diff
@@ -48,12 +48,25 @@ func PolicyReportResourceBuilder(pr *PolicyReport) *PolicyReportResource {
...
+	// Total number of policies in the report
+	node.Properties["numInsightPolicies"] = len(pr.Results)
```
jlpadilla, line 54: How will we use this field? Do we expect to use it for searching?  ie. `numInsightPolicies >2` ?

If it's only to display a total, then we could use length of `insightPolicies` instead.

SherinV: Do you mean using len(policies) instead of len(pr.Results)? @jlpadilla

jlpadilla: No, I meant that we could get the length **on the client side** instead of saving as a field.  It's just an alternative, it will depend on how we use the data.

The field `numInsightPolicies` will be searchable by users, so I wanted to confirm that is the expected behavior.

SherinV: @zlayne Can you comment on the usage of `numInsightPolicies`?

zlayne: This is a good question.. My original thought was to show the number of rule violations a PolicyReport contained (the `numInsightPolicies`). But I dont think searching by this field really has any value to the user, so maybe we just leave that transform to the search-ui?

SherinV: Since we are creating empty policyReports, if there are no violations, won't this field help in filtering out the empty ones? `numInsightPolicies > 0`

jlpadilla: ok, @SherinV use case sounds valid, we can use it in the suggested search.

jlpadilla (approved):
/lgtm
/approve

PR #127, opened 2021-05-03
add ownerName and ownerKind to node props

richiepiya:
Signed-off-by: Richie Piya <rpiyasirisilp@gmail.com>

**Related Issue:**  https://github.com/open-cluster-management/backlog/issues/9000

**Description of changes**
- get owner name and kind for pod resources

### DIFF
--- a/pkg/transforms/pod.go
+++ b/pkg/transforms/pod.go
@@ -106,6 +106,13 @@ func PodResourceBuilder(p *v1.Pod) *PodResource {
 
 	node := transformCommon(p) // Start off with the common properties
 
+	var ownerName = ""
+	var ownerKind = ""
+	if len(p.ObjectMeta.OwnerReferences) > 0 {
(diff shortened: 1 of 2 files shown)


SherinV:
Is this scenario not handled by the `ownedBy` edges?

richiepiya:
we need both the pod ownerName and ownerKind to match with resources that deploy these pods in app-ui

SherinV:
> we need both the pod ownerName and ownerKind to match with resources that deploy these pods in app-ui

Capturing our discussion here - currently, we have this info in search. Searching for the pod will bring up the replication-controller as a related resource and vice-versa. @richiepiya to check if this info can be used for the topology use-case.

richiepiya:
@SherinV updated the PR to conditionally pass the node properties

fyi @fxiang1

SherinV (approved):

PR #130, opened 2021-05-18
[release-2.0] Build selfLink when not present

jlpadilla:
**Related Issue:**  open-cluster-management/backlog#11904

### Description of changes
- Kubernetes 1.20 deprecates selfLink.  This adds a workaround that builds the selfLink to stay compatible with newer clusters.
- A permanent fix to remove our dependency on selfLink was merged to ACM 2.2 but is too big to back port.


pkg/transforms/transformer.go
```diff
@@ -82,6 +83,32 @@ type Deletion struct {
...
+		root = "apis"
+		groupVersion = "/" + nodeProps["apigroup"].(string) + "/" + nodeProps["apiversion"].(string)
+	} else if nodeProps["apigroup"] != nil && nodeProps["apigroup"].(string) != "" {
+		groupVersion = "/" + nodeProps["apigroup"].(string)
+	} else if nodeProps["apiversion"] != nil && nodeProps["apiversion"].(string) != "" {
+		groupVersion = "/" + nodeProps["apiversion"].(string)
+		root = "api"
+	}
+
+	selfLink := fmt.Sprintf("/%s%s%s/%s/%s", root, groupVersion, namespace, resourceString, nodeProps["name"].(string))
```
SherinV, line 107: Please add an example here to show how the string will be formatted.

pkg/transforms/transformer.go
```diff
@@ -82,6 +83,32 @@ type Deletion struct {
...
```
SherinV, line 107:
```suggestion
      // selfLink will be in the format "/api/v1/namespaces/default/pods/testpod"
	selfLink := fmt.Sprintf("/%s%s%s/%s/%s", root, groupVersion, namespace, resourceString, nodeProps["name"].(string))
```

pkg/transforms/transformer.go
```diff
@@ -92,6 +119,14 @@ func NewNodeEvent(event *Event, trans Transform, resourceString string) NodeEven
 		ComputeEdges: trans.BuildEdges,
 	}
 	ne.ResourceString = resourceString
+
+	// Kubernetes 1.20 deprecates selfLink. This workaround builds a self link to satisfy the existing logic.
+	// A permanent fix is delivered in ACM 2.2
+	if ne.Node.Properties["selfLink"] == nil || ne.Node.Properties["selfLink"].(string) == "" {
+		ne.Node.Properties["selfLink"] = buildSelfLink(ne.Node.Properties, resourceString)
+		fmt.Println("replacing selfLink")
```
SherinV, line 127: Remove the comment or replace with glog, if we are keeping it.

SherinV (approved):
lgtm other than minor comments.

pkg/transforms/transformer.go
jlpadilla: done

pkg/transforms/transformer.go
jlpadilla: done

zlayne (approved):

pkg/transforms/transformer.go
jlpadilla: Removed, this was for my debugging.

PR #132, opened 2021-05-19
Add search-collector lease addon on managed clusters

SherinV:
**Related Issue:**  open-cluster-management/backlog#6473

### Description of changes
Create and update search-collector lease on managed cluster instead of remote cluster namespace in hub.
Made necessary permission updates in klusterlet-addon-operator search chart. https://github.com/open-cluster-management/klusterlet-addon-operator/pull/21
Add test


pkg/config/kubeClient.go
```diff
@@ -53,9 +54,18 @@ func getKubeConfig() *rest.Config {
...
 func GetDiscoveryClient() *discovery.DiscoveryClient {
-	discoveryClient, err := discovery.NewDiscoveryClientForConfig(getKubeConfig())
+	discoveryClient, err := discovery.NewDiscoveryClientForConfig(GetKubeConfig())
 	if err != nil {
 		glog.Fatal("Cannot Construct Discovery Client From Config: ", err)
 	}
 	return discoveryClient
 }
+
+func GetKubeClient() *kubernetes.Clientset {
```
mqhackett, line 64: This was previously used code? Commented out and saved for later? It's finally later! Good move, we need this method now.

pkg/lease/lease.go
```diff
@@ -0,0 +1,82 @@
...
+		} else {
+			glog.Infof("addon lease %q/%q on local managed cluster updated", r.componentNamespace, r.LeaseName)
+		}
+		glog.Info("Exiting lease Reconcile")
+
+		return
+	}
+}
+
+func getPodNamespace() string {
```
mqhackett, line 75: The value of creating a separate package in our project is exemplified by this method. While it was previously recommended to put the lease creation logic in our operator, our current implementation disqualifies this option. It certainly would be possible, but would create a spaghetti code monster. 

By creating a separate function, we achieve functionality exposed by the operators that does not exist at the search-collector. Example, in the operator, we could just call a function that gives us the namespace, which is required for creating the lease. However, here, we need this method. 

Very clean way of doing this. I'd consider a name other than "lease" to avoid future confusion, but tbh i'm happy with 'lease' for now as it shouldn't cause any code-breaking collisions.

pkg/config/kubeClient.go
```diff
@@ -32,7 +33,7 @@ func GetDynamicClient() dynamic.Interface {
 	return dynamicClient
 }
 
-func getKubeConfig() *rest.Config {
+func GetKubeConfig() *rest.Config {
```
jlpadilla, line 36: Is this change needed? I don't see `GetKubeConfig()` being used from another package.

pkg/lease/lease.go
```diff
@@ -0,0 +1,73 @@
...
+				Namespace: r.componentNamespace,
+			},
+			Spec: coordinationv1.LeaseSpec{
+				LeaseDurationSeconds: &r.LeaseDurationSeconds,
+				RenewTime: &metav1.MicroTime{
+					Time: time.Now(),
+				},
+			},
+		}
+		if _, err := r.KubeClient.CoordinationV1().Leases(r.componentNamespace).Create(lease); err != nil {
```
jlpadilla, line 43: Most likely we'll need to update the pod permissions to do this create action.

jlpadilla: And the update action below.

jlpadilla (approved):
Left a couple minor comments.
/lgtm
/hold

pkg/lease/lease.go
SherinV: We updated the pod permissions for managed cluster collector pods in https://github.com/open-cluster-management/klusterlet-addon-operator/pull/21/files. Do we need to update the search-chart too? @jlpadilla

pkg/config/kubeClient.go
SherinV: Updated.

pkg/lease/lease.go
jlpadilla: @SherinV that's all we needed. Thanks for confirming! 👍

mqhackett: Could you explain why we needed to update the pod permissions for this call?

mqhackett (approved):

pkg/lease/lease.go
SherinV: So that the pod has create and update permissions on lease objects. Otherwise the create will fail with `pod doesn't have permission to create/update lease in this namespace`

PR #138, opened 2021-07-08
Handle when managed cluster does not support lease kind

SherinV:
**Related Issue:**  open-cluster-management/backlog#13558

### Description of changes
Create lease on remote cluster namespace on HUB if lease object is not available on the managed cluster (OCP 311)

### DIFF
--- a/main.go
+++ b/main.go
@@ -62,11 +62,13 @@ func main() {
 
 	if !config.Cfg.DeployedInHub {
 		leaseReconciler := lease.LeaseReconciler{
-			KubeClient:           config.GetKubeClient(),
+			HubKubeClient:        config.GetKubeClient(config.Cfg.AggregatorConfig),
+			KubeClient:           config.GetKubeClient(config.GetKubeConfig()),
 			LeaseName:            AddonName,
+			ClusterName:          config.Cfg.ClusterName,
 			LeaseDurationSeconds: int32(LeaseDurationSeconds),
 		}
-		glog.Info("Create/Update lease for search on managed cluster")
+		glog.Info("Create/Update lease for search")
 		go wait.Forever(leaseReconciler.Reconcile, time.Duration(leaseReconciler.LeaseDurationSeconds)*time.Second)
 	}
 
--- a/pkg/config/kubeClient.go
+++ b/pkg/config/kubeClient.go
@@ -70,7 +70,7 @@ func GetKubeClient(config *rest.Config) *kubernetes.Clientset {
 			glog.Fatal("Cannot Construct Kube Client from Config: ", err)
 		}
 	} else {
-	  glog.Error("Cannot Construct Kube Client as input Config is nil")
+		glog.Error("Cannot Construct Kube Client as input Config is nil")
 	}
 	return kubeClient
 }
--- a/pkg/config/kubeClient.go
+++ b/pkg/config/kubeClient.go
@@ -24,7 +24,7 @@ func GetDynamicClient() dynamic.Interface {
 	if dynamicClient != nil {
 		return dynamicClient
 	}
-	newDynamicClient, err := dynamic.NewForConfig(getKubeConfig())
+	newDynamicClient, err := dynamic.NewForConfig(GetKubeConfig())
 	if err != nil {
 		glog.Fatal("Cannot Construct Dynamic Client ", err)
 	}
@@ -33,7 +33,7 @@ func GetDynamicClient() dynamic.Interface {
 	return dynamicClient
 }
 
-func getKubeConfig() *rest.Config {
+func GetKubeConfig() *rest.Config {
 	var clientConfig *rest.Config
 	var clientConfigError error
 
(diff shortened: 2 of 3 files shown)


SherinV:
Similar to https://github.com/open-cluster-management/multicloud-operators-subscription/pull/516/files

main.go
```diff
@@ -62,11 +62,13 @@ func main() {
 
 	if !config.Cfg.DeployedInHub {
 		leaseReconciler := lease.LeaseReconciler{
-			KubeClient:           config.GetKubeClient(),
+			HubKubeClient:        config.GetKubeClient(config.Cfg.AggregatorConfig),
```
SherinV, line 65: @jlpadilla Is this the right config to use to connect to hub - to create lease if the managed cluster is OCP311?

elgnay:
The change looks fine with me. It's very similar to https://github.com/open-cluster-management-io/addon-framework/pull/23

pkg/lease/lease.go
```diff
@@ -13,25 +13,48 @@ import (
...
 	if len(r.componentNamespace) == 0 {
 		r.componentNamespace = getPodNamespace()
 	}
-	lease, err := r.KubeClient.CoordinationV1().Leases(r.componentNamespace).Get(r.LeaseName, metav1.GetOptions{})
+	// Create/update lease on managed cluster first. If it fails, it could mean lease resource kind
+	// is not supported on the managed cluster. Create/update lease on the hub then.
+	err := r.updateLease(r.componentNamespace, r.KubeClient)
+
+	if err != nil {
+		glog.Errorf("Failed to update lease %s/%s: %v on managed cluster", r.LeaseName, r.componentNamespace, err)
```
jlpadilla, line 33: We try to recover below, so don't log this error here.  It will be confusing if the recover logic succeeds.

pkg/lease/lease.go
```diff
@@ -13,25 +13,48 @@ import (
...
+
+			if err := r.updateLease(r.ClusterName, r.HubKubeClient); err != nil {
+				glog.Errorf("Failed to update lease %s/%s: %v on hub cluster", r.LeaseName, r.ClusterName, err)
+			}
+		}
+	}
+}
+
+func (r *LeaseReconciler) updateLease(namespace string, client kubernetes.Interface) error {
+	glog.Infof("Trying to update lease %q/%q", namespace, r.LeaseName)
```
jlpadilla, line 47: This will happen too often on a 3.1.1 cluster so log as `V(2)`

pkg/lease/lease.go
SherinV: I will move it into an else block - as we try to recover under certain conditions.

pkg/lease/lease.go
```diff
@@ -40,28 +63,31 @@ func (r *LeaseReconciler) Reconcile() {
...
-		} else {
-			glog.Infof("Addon lease %q/%q on managed cluster created for Search", r.componentNamespace, r.LeaseName)
+		if _, err := client.CoordinationV1().Leases(namespace).Create(lease); err != nil {
+			glog.Errorf("Unable to create addon lease %q/%q . error:%v", namespace, r.LeaseName, err)
+
+			return err
 		}
 
-		return
+		glog.Infof("Addon lease %q/%q created", namespace, r.LeaseName)
```
jlpadilla, line 72: Log as V(2)

pkg/lease/lease.go
```diff
@@ -40,28 +63,31 @@ func (r *LeaseReconciler) Reconcile() {
...
-		} else {
-			glog.V(2).Infof("Addon lease %q/%q on managed cluster updated for Search", r.componentNamespace, r.LeaseName)
+		if _, err = client.CoordinationV1().Leases(namespace).Update(lease); err != nil {
+			glog.Errorf("unable to update cluster lease %q/%q . error:%v", namespace, r.LeaseName, err)
+
+			return err
 		}
-		return
+
+		glog.Infof("addon lease %q/%q updated", namespace, r.LeaseName)
```
jlpadilla, line 88: reduce logging here too.

main.go
jlpadilla: I'm confused by the 2 different clients and config.  Let's chat.

pkg/config/kubeClient.go
```diff
@@ -54,17 +54,22 @@ func getKubeConfig() *rest.Config {
...
+func GetKubeClient(config *rest.Config) *kubernetes.Clientset {
+	var kubeClient *kubernetes.Clientset
+	var err error
+	if config != nil {
+		kubeClient, err = kubernetes.NewForConfig(config)
+		if err != nil {
+			glog.Fatal("Cannot Construct Kube Client from Config: ", err)
+		}
 	}
+	glog.Error("Cannot Construct Kube Client as input Config is nil")
```
jlpadilla, line 73: This needs to be inside an else or we'll always log it. 
```suggestion
	} else {
	  glog.Error("Cannot Construct Kube Client as input Config is nil")
	}
```

main.go
jlpadilla: Verified that the permissions of the hub kubeconfig are limited.
I wasn't allowed to create a lease in the default namespace.
```
E0708 19:17:11.545741       1 lease.go:80] Unable to get addon lease "default"/"search-collector" . error:leases.coordination.k8s.io "search-collector" is forbidden: User "system:open-cluster-management:cluster:jorge-managed:addon:search-collector:agent:zd9qb" cannot get resource "leases" in API group "coordination.k8s.io" in the namespace "default"
```

jlpadilla: Also validated that the lease can be created in the hub, but didn't test with OCP 3.11

pkg/config/kubeClient.go
jlpadilla: Committing because Sherin is unavailable.

jlpadilla (approved):
/lgtm

PR #140, opened 2021-07-08
PR to trigger build, there was a glitch after #138

jlpadilla:
**Related Issue:**  N/A

### Description of changes
- Go modules is on by default in 1.16


jlpadilla:
Not needed anymore. I did a manual release-ff into release-2.3 and it triggered the build and publish.

PR #191, opened 2022-03-10
Allowing users to allow/deny resources.

Anxhela21:

### Description of changes
This PR includes changes to search-collector so that it may look for the "allowdeny" ConfigMap and allow/deny specified resources per specified apiGroups.


go.mod
```diff
@@ -29,6 +29,8 @@ require (
 	sigs.k8s.io/application v0.8.3
 )
 
+require k8s.io/klog v1.0.0
```
jlpadilla, line 32: Our other projects are using `"k8s.io/klog/v2"`

pkg/informer/informer.go
```diff
@@ -30,6 +43,50 @@ type GenericInformer struct {
...
+
+type AllowedResources struct {
+	ApiGroups []string `yaml:"apiGroups"`
+	Resources []string `yaml:"resources"`
+}
+
+type DeniedResources struct {
+	ApiGroups []string `yaml:"apiGroups"`
+	Resources []string `yaml:"resources"`
+}
```
jlpadilla, line 71: These are the same, so make it generic.  Then when you create an instance the variable name will indicate allow or deny.

pkg/informer/informer.go
```diff
@@ -30,6 +43,50 @@ type GenericInformer struct {
...
+
+func GetAllowDenyData(cm *v1.ConfigMap) ([]AllowedResources, []DeniedResources) {
+
+	var allow []AllowedResources
+	allowerr := yaml.Unmarshal([]byte(cm.Data["AllowedResources"]), &allow)
+	if allowerr != nil {
+		klog.Fatalf("Unmarshal: %v", allowerr)
```
jlpadilla, line 78: `klog.Fatal()` will exit the go program.  Let's lower these to `klog.Error()`.  We should tolerate this error and continue with default values.

pkg/informer/informer.go
```diff
@@ -44,6 +101,154 @@ func InformerForResource(res schema.GroupVersionResource) (GenericInformer, erro
...
+
+	// fmt.Println(kind, group)
+	var boolVar bool
+
+	list := []string{"events", "projects", "clusters", "clusterstatuses", "oauthaccesstokens"}
+
+	//remove all apiResources with kind in list
+	for _, name := range list {
+		if kind == name {
+			fmt.Println("list of names, resource kind:", name, kind)
```
jlpadilla, line 114: Use `klog.V(5).Info`

pkg/informer/informer.go
```diff
@@ -44,6 +101,154 @@ func InformerForResource(res schema.GroupVersionResource) (GenericInformer, erro
...
+			boolVar = false
```
jlpadilla, line 115: Need to `return false` as soon as we hit this condition, otherwise the value of `boolVar` can get changed later.

pkg/informer/informer.go
```diff
@@ -44,6 +101,154 @@ func InformerForResource(res schema.GroupVersionResource) (GenericInformer, erro
 	return i, nil
 }
 
+func isResourceAllowed(cm *v1.ConfigMap, group, kind string, allowedList []AllowedResources, deniedList []DeniedResources) bool {
+
+	// fmt.Println(kind, group)
+	var boolVar bool
+
+	list := []string{"events", "projects", "clusters", "clusterstatuses", "oauthaccesstokens"}
```
jlpadilla, line 109: Please keep the comment from the original code that explains why we exclude these.

pkg/informer/informer.go
Anxhela21: will remove these comments used for debugging.

pkg/informer/informer.go
```diff
@@ -44,6 +101,154 @@ func InformerForResource(res schema.GroupVersionResource) (GenericInformer, erro
...
+		glog.Warning("ServerPreferredResources could not list all available resources: ", err)
+	}
+
+	// create client to get configmap
+	config := config.GetKubeConfig() //can't err here?
+	clientset, err := kubernetes.NewForConfig(config)
+
+	//locate the allow-deny ConfigMap:
+	var cm *v1.ConfigMap
+	if cm, err = clientset.CoreV1().ConfigMaps("open-cluster-management").Get(contextVar, "allowdeny-config", metav1.GetOptions{}); err != nil {
```
jlpadilla, line 199: We need to use the namespace where the collector pod is deployed.
We'll need to update the collector deployment to pass the namespace in an environment variable.  Here's an [example of the deployment.](https://github.com/stolostron/search-chart/blob/main/stable/search-prod/templates/operator-deployment.yaml#L51-L58)

pkg/informer/informer.go
```diff
@@ -44,6 +101,154 @@ func InformerForResource(res schema.GroupVersionResource) (GenericInformer, erro
...
+		glog.Warning("Can't find allow/deny ConfigMap", err)
+	}
+
+	var deniedList []DeniedResources
+	var allowedList []AllowedResources
+
+	//parse config:
+	allowedList, deniedList = GetAllowDenyData(cm)
```
jlpadilla, line 207: You can use `:=` instead of declaring the var.
```suggestion
	//parse config:
	allowedList, deniedList := GetAllowDenyData(cm)
```

pkg/informer/informer.go
Anxhela21: Okay, I will open a small PR to add that to collector deployment then do we just change the above to namespace to the env "WATCH_NAMESPACE"?

Anxhela21: Do we want to set env var to name as well?

pkg/informer/informer.go
```diff
@@ -44,6 +75,162 @@ func InformerForResource(res schema.GroupVersionResource) (GenericInformer, erro
...
+		}
+	}
+
+	// remove denied resources from all apigroups if * otherwise remove from specific apigroups.
+	for _, deny := range deniedList {
+		for _, api := range deny.ApiGroups {
+			for _, rec := range deny.Resources {
+				if api == "*" && rec != "*" {
+					if rec == kind {
+						klog.Info("deny all group, specific resource", group, api, kind, rec)
```
jlpadilla, line 102: Let's change the verbosity of the logs in this function `klog.V(1).Info(...)`

Message can be more descriptive.
```suggestion
						klog.V(1).Info("Excluding resource because it matches deny list. ", group, api, kind, rec)
```

jlpadilla: Or 
```
klog.V(1).Infof("Resource %s %s matched deny list. ", group, kind)
```

pkg/informer/informer.go
```diff
@@ -44,6 +75,162 @@ func InformerForResource(res schema.GroupVersionResource) (GenericInformer, erro
...
+							return true
+						}
+						// }
+					}
+				}
+			}
+		}
+	}
+
+	return boolVar
```
jlpadilla, line 162: - You don't need `boolVar`, just `return true` here.
- Let's add a log.Warning() before this return. Normally we should match one of the rules above and exit before this point, so logging a warning can help us notice if we missed any case.

pkg/informer/informer.go
```diff
@@ -44,6 +75,162 @@ func InformerForResource(res schema.GroupVersionResource) (GenericInformer, erro
 	return i, nil
 }
 
+func isResourceAllowed(cm *v1.ConfigMap, group, kind string, allowedList []Resource, deniedList []Resource) bool {
```
jlpadilla, line 78: I don't think we use `cm` in this function.
Also cm and allowList/denyList contain redundant information so let's not require both.

pkg/informer/informer.go
```diff
@@ -30,6 +39,28 @@ type GenericInformer struct {
...
+	ApiGroups []string `yaml:"apiGroups"`
+	Resources []string `yaml:"resources"`
+}
+
+func GetAllowDenyData(cm *v1.ConfigMap) ([]Resource, []Resource) {
+
+	var allow []Resource
+	allowerr := yaml.Unmarshal([]byte(cm.Data["AllowedResources"]), &allow)
+	if allowerr != nil {
+		klog.Errorf("Unmarshal: %v", allowerr)
```
jlpadilla, line 52: Let's be more descriptive with these 2 error messages to help the end user debug the problem.
```
klog.Errorf("Error while parsing allowed resources from ConfigMap. Can't use configured value, defaulting to allow all resources. %v", allowerr)
```

Anxhela21: Good point!

pkg/informer/informer.go
```diff
@@ -44,6 +75,160 @@ func InformerForResource(res schema.GroupVersionResource) (GenericInformer, erro
...
+	}
+
+	// create client to get configmap
+	config := config.GetKubeConfig() //can't err here?
+	clientset, err := kubernetes.NewForConfig(config)
+
+	//locate the allow-deny ConfigMap:
+	var cm *v1.ConfigMap
+	if cm, _ = clientset.CoreV1().ConfigMaps("POD_NAMESPACE").Get(contextVar, "allowdeny-config", metav1.GetOptions{}); err != nil {
+		glog.Warning("Can't find allowdeny-config ConfigMap", err)
```
jlpadilla, line 183: Not a warning because this is optional config, but lets make this very descriptive to help end users as much as possible.
```
glog.Info("Didn't find ConfigMap with name allowdeny-config. Will collect all resources. ", err)
```

pkg/informer/informer.go
```diff
@@ -44,6 +75,160 @@ func InformerForResource(res schema.GroupVersionResource) (GenericInformer, erro
...
+			for _, verb := range apiResource.Verbs {
+				if verb == "watch" {
+					watchResources = append(watchResources, apiResource)
+				}
+			}
+		}
+
+		watchList.APIResources = watchResources
+		supportedResources = append(supportedResources, &watchList)
+		// fmt.Println(supportedResources)
```
jlpadilla, line 222: delete

pkg/informer/informer.go
```diff
@@ -109,7 +294,7 @@ func (inform *GenericInformer) listAndResync() error {
 			return listError
 		}
 
-		// Add all resources.
+		// Add all resources. <-- this is where we should change to only add resources in Allow list:
```
jlpadilla, line 297: delete

pkg/informer/supportedResources.go
```diff
@@ -0,0 +1,192 @@
...
+
+	// create client to get configmap
+	config := config.GetKubeConfig() //can't err here?
+	clientset, err := kubernetes.NewForConfig(config)
+
+	//locate the allow-deny ConfigMap:
+	var cm *v1.ConfigMap
+	if cm, _ = clientset.CoreV1().ConfigMaps("POD_NAMESPACE").Get(contextVar, "search-collector-config", metav1.GetOptions{}); err != nil {
+		glog.Info("Didn't find ConfigMap with name allowdeny-config. Will collect all resources. ", err)
+	}
```
jlpadilla, line 147: - `err` still has the value from line 132 above.  Errors here are ignored because of `_`.

```go
	// create client to get configmap
	kubeClient := config.GetKubeClient(config.GetKubeConfig())

	//locate the search-collector-config ConfigMap:
	cm, err2 := kubeClient.CoreV1().ConfigMaps(config.Cfg.PodNamespace).Get(contextVar, "search-collector-config", metav1.GetOptions{})
	if err2 != nil {
		glog.Info("Didn't find ConfigMap with name search-collector-config. Will collect all resources. ", err2)
	}
```

pkg/config/config.go
```diff
@@ -31,6 +31,7 @@ const (
 	DEFAULT_AGGREGATOR_HOST    = "https://localhost"
 	DEFAULT_AGGREGATOR_PORT    = "3010"
 	DEFAULT_CLUSTER_NAME       = "local-cluster"
+	DEFAULT_POD_NAMESPACE      = "open-cluster-management"
```
mqhackett, line 34: Does anyone know when/if we will change the default namespace to stolostron?

pkg/config/config.go
```diff
@@ -40,21 +41,21 @@ const (
...
-	ClusterName          string `env:"CLUSTER_NAME"`       // The name of this cluster
-	ClusterNamespace     string `env:"CLUSTER_NAMESPACE"`  // The namespace of this cluster
-	DeployedInHub        bool   `env:"DEPLOYED_IN_HUB"`    // Tracks if deployed in the Hub or Managed cluster
-	HeartbeatMS          int    `env:"HEARTBEAT_MS"`       // Interval(ms) to send empty payload to ensure connection
-	KubeConfig           string `env:"KUBECONFIG"`         // Local kubeconfig path
-	MaxBackoffMS         int    `env:"MAX_BACKOFF_MS"`     // Maximum backoff in ms to wait after error
-	RediscoverRateMS     int    `env:"REDISCOVER_RATE_MS"` // Interval(ms) to poll for changes to CRDs
-	ReportRateMS         int    `env:"REPORT_RATE_MS"`     // Interval(ms) to send changes to the aggregator
-	RuntimeMode          string `env:"RUNTIME_MODE"`       // Running mode (development or production)
+	AggregatorConfig     *rest.Config // Config object for hub. Used to get TLS credentials.
```
mqhackett, line 44: What happened here? it looks like a little reformatting? I checked and it looks like all the letters are the same as before, so good.

pkg/informer/informer.go
```diff
@@ -179,7 +179,7 @@ func SupportedResources(discoveryClient *discovery.DiscoveryClient) (map[schema.
 
 	//locate the allow-deny ConfigMap:
 	var cm *v1.ConfigMap
-	if cm, err = clientset.CoreV1().ConfigMaps("POD_NAMESPACE").Get(contextVar, "allowdeny-config", metav1.GetOptions{}); err != nil {
+	if cm, _ = clientset.CoreV1().ConfigMaps("POD_NAMESPACE").Get(contextVar, "allowdeny-config", metav1.GetOptions{}); err != nil {
```
xrajesh: Not sure why we are removing err here .

Anxhela21: this is old @xrajesh, updated commits have the error.

Anxhela21: everything new is moved into it's own file `supportedResources.go`

pkg/config/config.go
Anxhela21: yes - should be same as before.

jlpadilla (approved):
/lgtm

Thank you @Anxhela21 !
Will add tests in separate PR.

PR #193, opened 2022-03-24
Test for allow/deny configmap

Anxhela21:

### Description of changes
test for allow/deny configmap

### DIFF
--- a/pkg/informer/sample-allowdeny.json
+++ b/pkg/informer/sample-allowdeny.json
@@ -1,0 +1,13 @@
+{
+    "apiVersion": "v1",
+    "data": {
+        "AllowedResources": "- apiGroups: \n    - \"*\"\n  resources: \n    - services\n    - pods\n- apiGroups:\n    - admission.k8s.io\n    - authentication.k8s.io\n  resources:\n    - \"*\"",
+        "DeniedResources": "- apiGroups:\n    - \"*\"\n  resources:\n    - secrets\n- apiGroups:\n    - admission.k8s.io\n  resources:\n    - policies\n    - iampolicies\n    - certificatepolicies"
+    },
+    "kind": "ConfigMap",
+    "metadata": {
+        "annotations": {},
+        "name": "search-collector-config",
+        "namespace": "open-cluster-management"
+    }
+}
--- a/pkg/informer/supportResources_test.go
+++ b/pkg/informer/supportResources_test.go
@@ -1,0 +1,135 @@
+package informer
+
+import (
+	// "fmt"
+
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	// "k8s.io/cli-runtime/pkg/genericclioptions"
+
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+//generate fakeclient to get configmap
+//
+func TestGetConfigMapByName(t *testing.T) {
+	//Test scenarios:
+	// Correct ConfigMap - clientset1
+	// No configmap. -
+	// Configmap without AllowResources or DenyResources keys.- clientset2
+	// Configmap with invalid yaml in AllowResources or DenyResources. clientset3
+	// Configmap with invalid Name - clientset
+	configmaps := []struct {
+		clientset kubernetes.Interface
+		n         string
+		ns        string
+		err       error
+	}{
+
+		{ // ConfigMap - Correct format for Allowed/Denied Resources and has correct Name
+			clientset: fake.NewSimpleClientset(&v1.ConfigMap{
+				TypeMeta: metav1.TypeMeta{
+					Kind:       "ConfigMap",
+					APIVersion: "v1",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "search-collector-config",
+					Namespace: "open-cluster-management",
+				},
+				Data: map[string]string{
+					"AllowedResources": "- apiGroups: \n    - \"*\"\n  resources: \n    - services\n    - pods\n- apiGroups:\n    - admission.k8s.io\n    - authentication.k8s.io\n  resources:\n    - \"*\"",
+					"DeniedResources":  "- apiGroups:\n    - \"*\"\n  resources:\n    - secrets\n- apiGroups:\n    - admission.k8s.io\n  resources:\n    - policies\n    - iampolicies\n    - certificatepolicies",
+				},
+			}), ns: "open-cluster-management",
+			n:   "search-collector-config",
+			err: nil},
+
+		{ // ConfigMap AllowedResources is missing key (first apiGroups)
+			clientset: fake.NewSimpleClientset(&v1.ConfigMap{
+				TypeMeta: metav1.TypeMeta{
+					Kind:       "ConfigMap",
+					APIVersion: "v1",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "search-collector-config",
+					Namespace: "open-cluster-management",
+				},
+				Data: map[string]string{
+					"AllowedResources": "- resources: \n    - services\n    - pods\n- apiGroups:\n    - admission.k8s.io\n    - authentication.k8s.io\n  resources:\n    - \"*\"",
+					"DeniedResources":  "- apiGroups:\n    - \"*\"\n  resources:\n    - secrets\n- apiGroups:\n    - admission.k8s.io\n  resources:\n    - policies\n    - iampolicies\n    - certificatepolicies",
+				},
+			}),
+			ns:  "open-cluster-management",
+			n:   "search-collector-config",
+			err: fmt.Errorf("This configMap is missing a key")},
+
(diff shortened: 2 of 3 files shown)


pkg/informer/sample-allowdeny.json
```diff
@@ -0,0 +1,13 @@
+{
```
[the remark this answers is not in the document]
Anxhela21: I was only using the json to get the string format for Allowed/DeniedResources. Would you prefer to have yaml fiels for each test case instead of configmap objects?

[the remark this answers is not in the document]
jlpadilla: no, the configmap objects are perfect.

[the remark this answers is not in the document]
Anxhela21: okay, we can delete this json file all together.

pkg/informer/supportResources_test.go
```diff
@@ -0,0 +1,135 @@
...
+func TestGetConfigMapByName(t *testing.T) {
+	//Test scenarios:
+	// Correct ConfigMap - clientset1
+	// No configmap. -
+	// Configmap without AllowResources or DenyResources keys.- clientset2
+	// Configmap with invalid yaml in AllowResources or DenyResources. clientset3
+	// Configmap with invalid Name - clientset
+	configmaps := []struct {
+		clientset kubernetes.Interface
+		n         string
```
mqhackett, line 31: Could we call these fields "name" and "namespace"? Personally, I think it makes the code more readable to have full words instead of abbreviations. It's more to type, but easier to interpret. 

Just a style preference. I'd be remiss if I didn't mention it.

pkg/informer/supportedResources.go
```diff
@@ -37,7 +37,7 @@ func GetAllowDenyData(cm *v1.ConfigMap) ([]Resource, []Resource) {
 		Can't use configured value, defaulting to deny all resources. %v`, denyerr)
 	}
 
-	return allow, deny
+	return allow, deny, allowerr, denyerr
```
SherinV, line 40: Why are we returning this error when we don't take any further action based on it? https://github.com/stolostron/search-collector/blob/f6ba1a73d94e169c1eeea99eca1df37d3eee339f/pkg/informer/supportedResources.go#L119

Anxhela21: We return these errors so that we can validate the errors are the same in the test case.

Anxhela21: (they are only used for the unit test)

pkg/informer/supportedResources.go
```diff
@@ -86,12 +86,48 @@ func isResourceAllowed(group, kind string, allowedList []Resource, deniedList []
...
+
+		for _, allow := range allowMap {
+			for _, deny := range denyMap {
+				if (allow["group"] == deny["group"] && allow["rec"] == deny["rec"]) || (allow["group"] == "*" && deny["group"] == "*" && allow["rec"] == deny["rec"]) ||
+					(allow["rec"] == "*" && deny["rec"] == "*" && allow["group"] == deny["group"]) {
+					glog.V(2).Infof("Deny Resource [group: '%s' kind: %s]. Resource present in both allow and deny rule.", group, kind)
+					return false
+				}
+			}
+		}
```
jlpadilla, line 102: This block of code is unreachable because this function returns when it matches the resource in the deny list above.

Anxhela21: ah, you are right! will refactor so it is included in both lists.

pkg/informer/supportedResources.go
```diff
@@ -37,11 +37,27 @@ func GetAllowDenyData(cm *v1.ConfigMap) ([]Resource, []Resource) {
...
+	for _, allow := range allowMap {
+		for _, deny := range denyMap {
+			if (allow["group"] == deny["group"] && allow["rec"] == deny["rec"]) ||
+				(allow["group"] == "*" && deny["group"] == "*" && allow["rec"] == deny["rec"]) ||
+				(allow["rec"] == "*" && deny["rec"] == "*" && allow["group"] == deny["group"]) {
+				glog.V(2).Infof("Deny Resource [group: '%s' kind: %s]. Resource present in both allow and deny rule.", group, kind)
+				return false
+			}
+		}
+	}
```
jlpadilla, line 59: How about using this logic instead....

1. Add a new function that checks if a resource matches a resource list (abstracting allow/deny).
```go
function isResourceMatchingList(resourceList Resource[], group, kind string) bool{
  for _, r := range resourceList {
    for _, g := range r.ApiGroups {
	  for _, k := range r.Resources {
              if (g == "*" || g == group) && (k == "*" || k == kind) { // Group and kind matches
	        return false
              }
	  }
      }
    }
}
```

2. Then rewrite the original logic using the new function.
```go
func isResourceAllowed(group, kind string, allowedList []Resource, deniedList []Resource) bool {
  // Ignore clusters and clusterstatus resources because these are handled by the aggregator.
  // Ignore oauthaccesstoken resources because those cause too much noise on OpenShift clusters.
  // Ignore projects as namespaces are overwritten to be projects on Openshift clusters - they tend to share
  // the same uid.
  list := []string{"events", "projects", "clusters", "clusterstatuses", "oauthaccesstokens"}
  // Deny all apiResources with kind in list
  for _, name := range list {
    if kind == name {
	    glog.V(2).Infof("Deny resource [group: '%s' kind: %s]. Search collector doesn't support it.", group, kind)
	    return false
    }
  }

  // Deny resources that match the deny list.
  if isResourceMatchingList(denyList, group, kind) {
     // Check if resource is also in the allow list.
     if isResourceMatchingList(allowList, group, kind) {
        glog.V(2).Infof("Deny Resource [group: '%s' kind: %s]. Resource present in both allow and deny rule.", group, kind)
     } else {
        glog.V(2).Infof("Deny resource [group: '%s' kind: %s]. Matched rule [group: '%s' kind: %s].", group, kind, g, k) 
     }
     return false
  }

  // If allowList not provided, interpret it as allow all resources.
  // otherwise allow only the resources declared in allow list.
  if len(allowedList) == 0 {
    glog.V(2).Infof("Allow resource [group: '%s' kind: %s]. AllowList is empty.", group, kind)
    return true
  } else {
    if isResourceMatchingList(allowList, group, kind) {
      glog.V(2).Infof("Allow resource [group: '%s' kind: %s]. Matched [group: '%s' kind: %s].", group, kind, g, k)
    return true
    }
  }
  glog.V(2).Infof("Deny resource [group: '%s' kind: %s]. It doesn't match any allow or deny rule.", group, kind)
  return false
}

Comment (Anxhela21) [reply pkg/informer/supportedResources.go:59]: Agree with this change. just minor change above we need to return the g and k from the isResourceMatchingList func.

File: pkg/informer/supportResources_test.go [hunk 6]
```diff
@@ -0,0 +1,216 @@
...
+
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	// "k8s.io/cli-runtime/pkg/genericclioptions"
```
jlpadilla, line 14: delete 2 commented imports.

pkg/informer/supportResources_test.go
```diff
@@ -0,0 +1,216 @@
...
+	incomingResources := make(map[string]string)
+	incomingGroup := make(map[string]string)
+
+	//testing for any resources:
+	incomingGroup["group"] = "authentication.k8s.io"
+	incomingResources["resource"] = "deployments"
+
+	//testing for any apigroup:
+	incomingGroup["group"] = "anygroup.io"
+	incomingResources["resource"] = "services"
```
jlpadilla, line 173: This overwrites the previous 2 lines.

pkg/informer/supportResources_test.go
```diff
@@ -0,0 +1,216 @@
...
+
+					t.Error("Not expected. Error", group, rec)
+				}
+			} else {
+				t.Error("Not expected. Error")
+			}
+
+		}
+
+	}
```
jlpadilla, line 189: I'd remove the map and just pass the values directly like this.

```go
// Test all resources allowed for group
allowed,_,_ := isResourceAllowed("authentication.k8s.io", "deployments", allow, deny)
if !allowed {
  t.Error("Expected resource deployments.authentication.k8s.io to be allowed, but got denied.")
}

// Test resource matching any group
allowed,_,_ := isResourceAllowed("anygroup.io", "services", allow, deny)
if !allowed {
  t.Error("Expected resource services.anygroup.io to be allowed, but got denied.")
}

// Continue the same pattern with other tests.
}

Comment (jlpadilla) [reply pkg/informer/supportResources_test.go:189]: Wait, the above makes it clear that we are testing `isResourceAllowed()` with different inputs, but it creates a fair amount of code duplication.  Let me try to refactor again...

Comment (Anxhela21) [reply pkg/informer/supportResources_test.go:189]: that's much simpler!

Comment (jlpadilla) [reply pkg/informer/supportResources_test.go:189]: How about this?
```go
testData := [][]string{{"authentication.k8s.io", "deployments"}, {"services.anygroup.io", "services"}}

for data := range testData {
  allowed := isResourceAllowed(data[0], data[1], allow, deny)
  if !allowed {
    t.Error("Expected group: %s resource: %s to be allowed, but got denied.", data[0], data[1])
  }
}
```

jlpadilla (approved):
/lgtm
/hold

PR #195, opened 2022-04-08
[2.4] Filter grc source violations from policyreport

zlayne:
Signed-off-by: Zack Layne <zlayne@redhat.com>

### Description of changes
Filter grc sourced violations from PolicyReport search responses.


xrajesh:
@zlayne Looks good , but the test case yaml , should it have the source:grc also to verify , we pick up only insights?

zlayne:
@xrajesh yea that'd be a bit more robust.

xrajesh (approved):

crizzo71:
@smithbw88 @evelinec Adding this to 2.4.3 RC2 -

PR #197, opened 2022-04-14
[Release-2.4] allow/deny resources feature for 2.4

Anxhela21:

### Description of changes
Adding this feature for search-collector 2.4

As an ACM user, I want to control the resources collected by search from my managed clusters

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ lint:
 	GOGC=25 golangci-lint run --timeout=3m
 
 run:
-	GOGC=25 go run main.go
+	GOGC=25 go run main.go --v=2
 
 .PHONY: test
 test:
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -26,7 +26,7 @@ import (
 
 // Out of box defaults
 const (
-	COLLECTOR_API_VERSION      = "2.5.0"
+	COLLECTOR_API_VERSION      = "2.4.0"
 	DEFAULT_AGGREGATOR_URL     = "https://localhost:3010" // this will be deprecated in the future
 	DEFAULT_AGGREGATOR_HOST    = "https://localhost"
 	DEFAULT_AGGREGATOR_PORT    = "3010"
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -26,11 +26,12 @@ import (
 
 // Out of box defaults
 const (
-	COLLECTOR_API_VERSION      = "2.4.0"
+	COLLECTOR_API_VERSION      = "2.5.0"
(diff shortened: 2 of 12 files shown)


Makefile
```diff
@@ -34,7 +34,7 @@ lint:
 	GOGC=25 golangci-lint run --timeout=3m
 
 run:
-	GOGC=25 go run main.go
+	GOGC=25 go run main.go --v=2
```
mqhackett, line 37: Why do we turn up the verbosity here?

Anxhela21: When we run `make run` without `--v=2`, collector hides the groups/resources it finds. This was mostly useful when testing allow/deny list. However, I'm not sure if we want to keep this change so those logs are always visible. Thoughts? @jlpadilla

jlpadilla: This line only affects development.  It's ok to develop with a higher verbosity if that helps. I have set a similar default in other projects.

This doesn't affect production, but just FYI, in prod we should try to reduce the default logging to only a summary, so that problems and potential problems stand out.

pkg/config/config.go
```diff
@@ -26,11 +26,12 @@ import (
 
 // Out of box defaults
 const (
-	COLLECTOR_API_VERSION      = "2.4.0"
+	COLLECTOR_API_VERSION      = "2.5.0"
```
SherinV, line 29: Should this be updated?

pkg/informer/supportedResources.go
```diff
@@ -0,0 +1,173 @@
+// Copyright (c) 2020 Red Hat, Inc.
```
SherinV, line 1: Remove this line

SherinV (commented):
Added a few minor comments.

pkg/informer/supportedResources.go
Anxhela21: Every script in this project has that in the header.

pkg/config/config.go
SherinV: I believe it should remain 2.4, in release 2.4.

Anxhela21: nope, will change that back to 2.4. thanks!

pkg/informer/supportedResources.go
SherinV: We used to add that before. But after we open-sourced, the recommendation is not to have that. Since, this is a new file it should only have  

> // Copyright Contributors to the Open Cluster Management project

Anxhela21: Gotcha, will change this file here and open PR for the others in this repo.

SherinV (approved):

PR #224, opened 2022-10-12
Watch for search-collector-config

Anxhela21:

**Related Issue:**  [Jira link to issue](https://issues.redhat.com/secure/RapidBoard.jspa?rapidView=15705&view=detail&selectedIssue=ACM-1804)

### Description of changes
We need Search Configurable Collection feature to have GA quality. The collector should watch and accurately react to changes user makes on the configmap. The documentation should also move out of readme and into product doc.


main.go
```diff
@@ -158,6 +158,10 @@ func main() {
 				// Now, loop through the new list, which after the above deletions, contains only stuff that needs to
 				// have a new informer created for it.
 				for gvr := range gvrList {
+					fmt.Println("Found new resource %s, creating informer\n", gvr.String())
+					if gvr.Resource == "configmaps" {
+						fmt.Println("FOUND CONFIGMAP RESOURCE", gvr.String())
+					}
 					glog.V(2).Infof("Found new resource %s, creating informer\n", gvr.String())
```
jlpadilla, line 165: This path should work now.  When a new resource is allowed I see this log (after a few minutes) and the new resource gets collected.

The problem is when we remove a resource that's being watched.  We stop the informer, but don't remove the data. The informer gets stopped a few lines above here, in this block.
```go
glog.V(2).Infof("Resource %s no longer exists or no longer supports watch, stopping its informer\n", gvr.String())
close(stopper)
delete(stoppers, gvr)
```

**A better starting point is where the informer exits its execution, around this line**. https://github.com/stolostron/search-collector/blob/main/pkg/informer/informer.go#L51-L53

The informer needs to trigger a delete event for each resource in the `resourceIndex` before it terminates.

README.md
```diff
@@ -56,56 +56,6 @@ RUNTIME_MODE       | no       | production               | Running mode (develop
 - Environment variables can also be set in the `./config.json` for development. If both provide a value for a specific property, the environment variable overrides the file. You can define your own `config.json` file and pass it to the application with the following command: `-c <config_file>`
 - The application can take any flags for [glog](https://github.com/golang/glog), which passes them straight into glog. The glog flag `--logtostderr` is set to true by default.
 
-### Dev Preview (Search Configurable Collection)
```
jlpadilla, line 59: Where is this topic moving to?
To avoid breaking existing links, let's add a pointer to the new document from here. We also need to keep the title unchanged.

```
### Dev Preview (Search Configurable Collection)
Configurable collection is now fully supported. This topic has moved to <add link here>.
```

Anxhela21: I was thinking we can add to wiki within search-collector for now, unless there is a better place for it. okay, will keep the title and link to wiki until the product doc is ready.

Makefile
```diff
@@ -4,7 +4,7 @@
 include build/Configfile
 BINDIR ?= output
 
-USE_VENDORIZED_BUILD_HARNESS ?=
+USE_VENDORIZED_BUILD_HARNESS ?= true
```
jlpadilla, line 7: Please remove this change from the PR.

README.md
```diff
@@ -1,18 +1,23 @@
...
 
-See the [Data model documentation](https://github.com/stolostron/search-collector/blob/pkg/transforms/README.md) for more information.
+See the [Data model documentation](https://github.com/stolostron/search-collector/blob/main/pkg/transforms/README.md) for more information.
+
+## Features
+
+Search-collector provides a similar solution as Kubernetes controllers and consists of four main components.
+ 
+* **Informer**: queries for resources and watches for updates
+* **Transformer**: applies updates to resources
```
jlpadilla, line 18:
```suggestion
* **Transformer**: extracts data from resources to include in the search index. Discovers relationships to other resources.
```

README.md
```diff
@@ -1,18 +1,23 @@
...
+* **Sender**: sends quachanges to the search-indexer
+* **Reconciler**: merges changes to search-indexer
```
jlpadilla, line 20:
```suggestion
* **Reconciler**: merges changes into search-collector internal state.
```

README.md
jlpadilla: Yes, the search-v2-operator wiki works.

README.md
```diff
@@ -1,18 +1,23 @@
 [comment]: # ( Copyright Contributors to the Open Cluster Management project )
 
-# WORK IN PROGRESS
-
-We are in the process of enabling this repo for community contribution. See wiki [here](https://open-cluster-management.io/concepts/architecture/).
-
 # Open cluster management - Search collector
 
 The `search-collector` is part of the search component in Open Cluster Management. The [search feature spec](https://github.com/stolostron/search/blob/main/feature-spec/search.md) has an overview of all the search components.
```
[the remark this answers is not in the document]
Anxhela21: Done.

pkg/informer/informer_test.go
```diff
@@ -126,6 +126,10 @@ func Test_listAndResync_syncWithPrevState(t *testing.T) {
 	if *deleteFuncCount != 1 {
 		t.Errorf("Expected informer.DeleteFunc to be called 1 time, but got %d.", *deleteFuncCount)
 	}
+	// Verify the resources no longer exist in informer.resourceIndex:
+	if _, ok := informer.resourceIndex["fake-uid"]; ok {
+		t.Errorf("Expected informer.resourceIndex to be empty")
+	}
```
jlpadilla, line 132: A better test would be to validate that `inform.DeleteFunc` gets called with the correct args.

jlpadilla: Actually, we need a new test where we stop the informer and validate that  `inform.DeleteFunc` is called with the expected data.

pkg/informer/informer.go
```diff
@@ -50,6 +50,15 @@ func (inform *GenericInformer) Run(stopper chan struct{}) {
...
+
+			for key := range inform.resourceIndex {
+				glog.V(5).Infof("Stopping informer %s and removing resource with UID: %s", inform.gvr.Resource, key)
+				obj := newUnstructured(inform.gvr.Resource, key)
+				inform.DeleteFunc(obj)
+				glog.V(5).Info("Informer stopped. ", inform.gvr.String())
+
+			}
+
 			return
```
jlpadilla, line 62: Move log to be more accurate and reduce duplicate information.
```suggestion
			for key := range inform.resourceIndex {
				glog.V(5).Infof("Stopping informer %s and removing resource with UID: %s", inform.gvr.Resource, key)
				obj := newUnstructured(inform.gvr.Resource, key)
				inform.DeleteFunc(obj)
			}
			glog.Info("Informer stopped. ", inform.gvr.String())
			return
```

pkg/informer/informer_test.go
```diff
@@ -126,6 +126,50 @@ func Test_listAndResync_syncWithPrevState(t *testing.T) {
...
+	informer.resourceIndex["id-100"] = "54321"
+
+	//keep track of calls made to mock informer.DeleteFunc below
+	deleteFuncCalls := make([]string, 0)
+
+	// mock DeleteFunc and get uids
+	informer.DeleteFunc = func(obj interface{}) {
+		for key := range informer.resourceIndex {
+			obj := newUnstructured(informer.gvr.Resource, key)
+			uid = string(obj.GetUID())
```
jlpadilla, line 149: Remove `var uid string` and use `:=` instead.
```suggestion
			uid := string(obj.GetUID())
```

pkg/informer/informer_test.go
```diff
@@ -126,6 +126,50 @@ func Test_listAndResync_syncWithPrevState(t *testing.T) {
...
+
+func Test_StoppedInformer_ValidateDeleteFunc(t *testing.T) {
+	var uid string
+
+	//create informer for mock resource
+	informer, _, _, _ := initInformer()
+
+	//mock informer state
+	informer.resourceIndex["id-999"] = "12345"
+	informer.resourceIndex["id-100"] = "54321"
```
jlpadilla, line 140: Declare these in-line.  The difference isn't too noticeable here, but I'm trying to use in-line declaration more consistently because it's easier to visualize the object that we are mocking, especially when it has more data.

```suggestion
	informer.resourceIndex = map[string]string{"id-999": "12345", "id-100": "54321"}
```

pkg/informer/informer_test.go
```diff
@@ -126,6 +126,50 @@ func Test_listAndResync_syncWithPrevState(t *testing.T) {
...
+	time.Sleep(10 * time.Millisecond)
+
+	//exist informer to trigger DeleteFunc
+	close(stopper)
+
+	//allow test to process
+	time.Sleep(10 * time.Millisecond)
+
+	// Verify that the informer.DeleteFunc was called with uid=id-999 and uid=id-100
+	valid := map[string]bool{"id-999": true, "id-100": true}
```
jlpadilla, line 166: This should match the `inform.resourceIndex`.  So you can:
1. Change to `map[string]string`
2. declare this object at the beginning of the test and use it to create the mock.
  ```
  resources := map[string]string{"id-999": "12345", "id-100": "54321"}
  //
  //
  informer.resourceIndex = resources
  ```

PR #226, opened 2022-11-01
Collect resources and conditions from ArgoCD application status

xiangjingli:

https://issues.redhat.com/browse/ACM-1702

### Description of changes
enhance search collectors to be able to collect resource list and error/warning conditions from the argocd application status

### DIFF
--- a/pkg/transforms/argoapplication.go
+++ b/pkg/transforms/argoapplication.go
@@ -5,12 +5,14 @@ package transforms
 import (
 	"strings"
 
+	"github.com/golang/glog"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // ArgoApplicationResource ...
 type ArgoApplicationResource struct {
-	node Node
+	node      Node
+	resources []ResourceStatus
 }
 
 type ArgoApplication struct {
@@ -39,11 +41,31 @@ type ArgoApplicationDestination struct {
 }
 
 type ArgoApplicationStatus struct {
-	Health ArgoApplicationHealth `json:"health" protobuf:"bytes,1,opt,name=health"`
+	Resources  []ResourceStatus       `json:"resources,omitempty" protobuf:"bytes,1,opt,name=resources"`
+	Conditions []ApplicationCondition `json:"conditions,omitempty" protobuf:"bytes,5,opt,name=conditions"`
+	Health     HealthStatus           `json:"health" protobuf:"bytes,1,opt,name=health"`
+	Sync       SyncStatus             `json:"sync,omitempty" protobuf:"bytes,2,opt,name=sync"`
 }
 
-type ArgoApplicationHealth struct {
+type ResourceStatus struct {
+	Group     string `json:"group,omitempty" protobuf:"bytes,1,opt,name=group"`
+	Version   string `json:"version,omitempty" protobuf:"bytes,2,opt,name=version"`
+	Kind      string `json:"kind,omitempty" protobuf:"bytes,3,opt,name=kind"`
+	Namespace string `json:"namespace,omitempty" protobuf:"bytes,4,opt,name=namespace"`
+	Name      string `json:"name,omitempty" protobuf:"bytes,5,opt,name=name"`
+}
+
+type ApplicationCondition struct {
+	Type    string `json:"type" protobuf:"bytes,1,opt,name=type"`
+	Message string `json:"message" protobuf:"bytes,2,opt,name=message"`
+}
+
+type HealthStatus struct {
 	Status string `json:"status" protobuf:"bytes,1,opt,name=status"`
+}
+
+type SyncStatus struct {
+	Status string `json:"status" protobuf:"bytes,1,opt,name=status,casttype=SyncStatusCode"`
 }
 
 // ArgoApplicationResourceBuilder ...
@@ -53,14 +75,17 @@ func ArgoApplicationResourceBuilder(a *ArgoApplication) *ArgoApplicationResource
 
 	// Extract the properties specific to this type
 
-	// ApplicationSet
+	// add its ApplicationSet owner
 	applicationSet := ""
+
 	for _, ref := range a.ObjectMeta.OwnerReferences {
 		if strings.Index(ref.APIVersion, "argoproj.io/") == 0 && ref.Kind == "ApplicationSet" {
 			applicationSet = ref.Name
+
 			break
 		}
 	}
+
 	node.Properties["applicationSet"] = applicationSet
 
 	// Destination properties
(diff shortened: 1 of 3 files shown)


xiangjingli:
/assign @jlpadilla  @fxiang1 

Could you help review the PR when you have a cycle?  

- The PR is for collecting the resource list and the condition list. Since both are located in the ArgoCD application status, I implemented it in node build function instead of edge build function. Edge list is for saving other resources related to the original node (owned, referred etc) and it seems that Edge can't save much information except for uid and kind

- The resource list are collected without truncation as we don't want to miss any resource deployed by the ArgoCD application. 
- The condition list are collected with message truncation. The size of each message is up to 512 byte because the argocd error message normally is pretty long.

fxiang1 (approved):
LGTM! Thanks Xiangjing!

pkg/transforms/argoapplication.go
```diff
@@ -73,16 +98,61 @@ func ArgoApplicationResourceBuilder(a *ArgoApplication) *ArgoApplicationResource
...
+		}
+
+		truncatedConditions = append(truncatedConditions, newCondition)
+	}
+
+	truncatedConditionsByte, err := yaml.Marshal(truncatedConditions)
+	if err != nil {
+		glog.Errorf("Marshaling ArgoCD application %v/%v condition list failed: %v", a.Namespace, a.Name, err)
+	} else {
+		node.Properties["conditions"] = string(truncatedConditionsByte)
```
jlpadilla, line 134: This is overloading too much data into the `conditions` property. This will make it difficult to search for a particular condition.

- **Option 1:** We could use a `[]string` where each condition is a separate item. It will behave similar to `label`. This is better but still not ideal, especially if a condition can have multiple messages.
- **Option 2:** We could use the `condition.Type` as the property key.

pkg/transforms/argoapplication.go
```diff
@@ -73,16 +98,61 @@ func ArgoApplicationResourceBuilder(a *ArgoApplication) *ArgoApplicationResource
...
+	}
 
 	return &ArgoApplicationResource{node: node}
 }
 
+func TruncateText(text string, width int) string {
+	if width < 0 {
+		return ""
+	}
```
jlpadilla, line 143: Probably we should log a warning if this condition is true.

pkg/transforms/argoapplication.go
```diff
@@ -73,16 +98,61 @@ func ArgoApplicationResourceBuilder(a *ArgoApplication) *ArgoApplicationResource
...
-	node.Properties["status"] = a.Status.Health.Status
+	node.Properties["healthStatus"] = a.Status.Health.Status
+	node.Properties["syncStatus"] = a.Status.Sync.Status
+
+	// resource list properties
+	resourcesByte, err := yaml.Marshal(a.Status.Resources)
+	if err != nil {
+		glog.Errorf("Marshaling ArgoCD application %v/%v resource list failed: %v", a.Namespace, a.Name, err)
+	} else {
+		node.Properties["resources"] = string(resourcesByte)
```
jlpadilla, line 115: Same as comment https://github.com/stolostron/search-collector/pull/226/files#r1013266422

The `resources` property is encoding too much information, which will make the search functionality difficult.

In this case, I'd prefer to implement this as an edge to the resource. This will fit better with the relationship model used by search and avoid duplicate information.

pkg/transforms/argoapplication.go
```diff
@@ -73,16 +98,61 @@ func ArgoApplicationResourceBuilder(a *ArgoApplication) *ArgoApplicationResource
...
 
 	return &ArgoApplicationResource{node: node}
 }
 
+func TruncateText(text string, width int) string {
+	if width < 0 {
+		return ""
+	}
+
+	res := []rune(text)
```
SherinV, line 145: Will this work?

```suggestion

if width < 0 {
		return ""
	}
	if len(text) > width {
		return text[:width] + "..."
	}
	return text
```

test-data/argoapplication.json
```diff
@@ -44,6 +44,46 @@
...
+    ],
+    "conditions": [
+        {
+            "type": "SyncError",
+            "message": "serviceaccounts is forbidden: User \"system:serviceaccount:openshift-gitops:openshift-gitops-argocd-application-controller\" cannot create resource \"serviceaccounts\" in API group \"\" in the namespace \"helm-nginx\""
+        },
+        {
+            "type": "InvalidSpecError",
+            "message": "invalid spec in the argocd application"
+        }
```
jlpadilla, line 85: Can you add an example where the message is larger than 512 characters to verify the truncation logic?

pkg/transforms/argoapplication.go
```diff
@@ -92,5 +162,6 @@ func (a ArgoApplicationResource) BuildNode() Node {
 // See documentation at pkg/transforms/README.md
 func (a ArgoApplicationResource) BuildEdges(ns NodeStore) []Edge {
 	ret := []Edge{}
+
 	return ret
```
SherinV, line 166:
```suggestion
	return []Edge{}
```

pkg/transforms/argoapplication.go
```diff
@@ -73,16 +98,61 @@ func ArgoApplicationResourceBuilder(a *ArgoApplication) *ArgoApplicationResource
...
+
+	res := []rune(text)
+
+	if len(res) > width {
+		res = res[:width]
+
+		return string(res) + "..."
+	}
+
+	return string(res)
```
SherinV, line 153: Will this work for you?
```suggestion
		if len(text) > width {
		return text[:width] + "..."
	}
	return text
```

SherinV (dismissed):
Please add the commit on TruncateText function if that change works.

pkg/transforms/argoapplication.go
SherinV: ok.

pkg/transforms/argoapplication.go
```diff
@@ -73,14 +98,43 @@ func ArgoApplicationResourceBuilder(a *ArgoApplication) *ArgoApplicationResource
...
+	node.Properties["healthStatus"] = a.Status.Health.Status
+	node.Properties["syncStatus"] = a.Status.Sync.Status
+
+	// conditions properties, each condition type is a property
+	for _, condition := range a.Status.Conditions {
+		truncatedMessage := TruncateText(condition.Message, 512)
+		if truncatedMessage > "" {
+			node.Properties["condition"+condition.Type] = truncatedMessage
+		}
+	}
```
jlpadilla, line 116: Status conditions are generic to all resources, so I want to take a closer look at the forma so it's useful in the context of searching for resources matching a given condition.

#### Sample conditions for a Pod resource.
image

jlpadilla: I don't want to block this PR either. Maybe we could use a temporary property name like this?
```go
node.Properties["_internalCondition"+condition.Type] = truncatedMessage
```

xiangjingli: @jlpadilla  Thanks! just updated the PR to the the temporary property name.  We can create a new PR once you figure it out.

jlpadilla: Thank you! Hopefully we can revisit this soon.

jlpadilla:
Follow up story to collect status conditions from any resource. https://issues.redhat.com/browse/ACM-2071

PR #227, opened 2022-12-12
add hosting application set property in support of argocd integration…

xiangjingli:
… with pull model

### Description of changes

1. Add a new property `hostingApplicationSet` to indicate the hosting applicationset if the argocd application is deployed by pull model. 
The `hostingApplicationSet` info is from the hosting-applicationset annotation in the argocd application CR on the managed cluster 

2. While the existing `applicationSet` is to show the its parent applicationset if the argocd app is deployed by push model
The applicationSet info is from the owner reference in the argocd application CR on the hub cluster.

To-Do:
We will work with app UI in 2.8 to see if we can use one applicationSet field to show the parent argocd applicationset, regardless of the argo application being deployed by pull model or push model.

### DIFF
--- a/pkg/transforms/argoapplication.go
+++ b/pkg/transforms/argoapplication.go
@@ -88,6 +88,13 @@ func ArgoApplicationResourceBuilder(a *ArgoApplication) *ArgoApplicationResource
 
 	node.Properties["applicationSet"] = applicationSet
 
+	// add its hosting applicationSet namespaced name, if the argocd app is propogated by ACM argocd pull integration controller
+	if a.Annotations != nil {
+		if a.Annotations["apps.open-cluster-management.io/hosting-applicationset"] != "" {
+			node.Properties["_hostingResource"] = "ApplicationSet/" + a.Annotations["apps.open-cluster-management.io/hosting-applicationset"]
+		}
+	}
+
 	// Destination properties
 	node.Properties["destinationName"] = a.Spec.Destination.Name
 	node.Properties["destinationNamespace"] = a.Spec.Destination.Namespace
--- a/pkg/transforms/argoapplication_test.go
+++ b/pkg/transforms/argoapplication_test.go
@@ -16,6 +16,7 @@ func TestTransformArgoApplication(t *testing.T) {
 	// Test only the fields that exist in application - the common test will test the other bits
 	AssertEqual("kind", node.Properties["kind"], "Application", t)
 	AssertEqual("applicationSet", node.Properties["applicationSet"], "helloworld-set", t)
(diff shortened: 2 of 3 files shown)


pkg/transforms/argoapplication.go
```diff
@@ -88,6 +88,17 @@ func ArgoApplicationResourceBuilder(a *ArgoApplication) *ArgoApplicationResource
...
+	hostingApplicationSet := ""
+
+	if a.Annotations != nil {
+		if a.Annotations["apps.open-cluster-management.io/hosting-applicationset"] != "" {
+			hostingApplicationSet = a.Annotations["apps.open-cluster-management.io/hosting-applicationset"]
+		}
+	}
+
+	node.Properties["hostingApplicationSet"] = hostingApplicationSet
+
```
jlpadilla, line 101: Can we only add the property if not empty?

```suggestion
	// add its hosting applicationSet namespaced name, if the argocd app is propogated by ACM argocd pull integration controller

	if a.Annotations != nil {
		if a.Annotations["apps.open-cluster-management.io/hosting-applicationset"] != "" {
			node.Properties["hostingApplicationSet"] = a.Annotations["apps.open-cluster-management.io/hosting-applicationset"]
		}
	}

```

jlpadilla: Also could we make this property internal by adding `_` to the name?  `_hostingApplicationSet"

xiangjingli: fixed

pkg/transforms/argoapplication.go
```diff
@@ -88,6 +88,13 @@ func ArgoApplicationResourceBuilder(a *ArgoApplication) *ArgoApplicationResource
 
 	node.Properties["applicationSet"] = applicationSet
 
+	// add its hosting applicationSet namespaced name, if the argocd app is propogated by ACM argocd pull integration controller
+	if a.Annotations != nil {
+		if a.Annotations["apps.open-cluster-management.io/hosting-applicationset"] != "" {
+			node.Properties["_hostingResource"] = "ApplicationSet/" + a.Annotations["apps.open-cluster-management.io/hosting-applicationset"]
```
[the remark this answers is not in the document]
SherinV: It should work. We will need to make code changes where we refer to this property to use the new format, including the [trigger in the postgres database](https://github.com/stolostron/search-v2-operator/blob/8df23cefcfcbaaa7a30bbd5c18fdde86e080f9c0/controllers/create_pgconfigmap.go#L48).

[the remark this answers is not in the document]
jlpadilla: Thank you! 
The additional changes will be for 2.8

jlpadilla (approved):
/lgtm

PR #228, opened 2022-12-15
append operator error message to property

xiangjingli:

### Description of changes

append the error message is in operationState.message to the error condition property

### DIFF
--- a/pkg/transforms/argoapplication.go
+++ b/pkg/transforms/argoapplication.go
@@ -41,10 +41,11 @@ type ArgoApplicationDestination struct {
 }
 
 type ArgoApplicationStatus struct {
-	Resources  []ResourceStatus       `json:"resources,omitempty" protobuf:"bytes,1,opt,name=resources"`
-	Conditions []ApplicationCondition `json:"conditions,omitempty" protobuf:"bytes,5,opt,name=conditions"`
-	Health     HealthStatus           `json:"health" protobuf:"bytes,1,opt,name=health"`
-	Sync       SyncStatus             `json:"sync,omitempty" protobuf:"bytes,2,opt,name=sync"`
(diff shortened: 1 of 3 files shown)


pkg/transforms/argoapplication.go
```diff
@@ -122,6 +129,16 @@ func ArgoApplicationResourceBuilder(a *ArgoApplication) *ArgoApplicationResource
 		}
 	}
 
+	// if there is operationState.message, append it to the error condition property
+	if a.Status.Health.Status != "Healthy" || a.Status.Sync.Status != "Synced" {
+		if a.Status.OperationState != nil && a.Status.OperationState.Message != "" {
+			truncatedMessage := TruncateText(a.Status.OperationState.Message, 512)
+			if truncatedMessage > "" {
+				node.Properties["_internalCondition"+"OperationError"] = truncatedMessage
```
jlpadilla, line 137: Maybe the name of this property should be `_operationStateMessage`?
The `_` means that it's an internal property, so you can omit the word "internal".  
Could you also remove "internal" from the condition above?

xiangjingli: @jlpadilla  Here is what I updated
- removed the `internal` keyword. So the condition property naming convention is `_condition<type>`  

- The <type> is from argocd status condition type, its naming convention is <error type> with the suffix Error or Warning
https://github.com/argoproj/argo-cd/blob/22ea51d4e40f4a57d825eeb8c0f483bc7150d340/pkg/apis/application/v1alpha1/types.go#L963-L980

- The error type for the operationState.message is called `_conditionOperationError` to keep the consistency with other Argocd app condition types.

jlpadilla (approved):
/lgtm

PR #230, opened 2022-12-23
don't bundle operationState message if its phase is succeeded

xiangjingli:

### Description of changes

We found that one argo application status could be like this
- its health status is not healthy 
- its sync status is  not synced
- while its operationState phase is succeeded 

In this case, we should not bundle the operationState message as it is not error message


jlpadilla (approved):
/lgtm

Merging because this is a bug, but keep in mind that 2.7 is on feature complete. Thanks!

PR #242, opened 2023-05-11
Add jitter to backoff logic

jlpadilla:
### Related Issue
N/A

### Description of changes
- Adds random jitter to the backoff logic to prevent collectors form getting in the same retry cycle.


bjoydeep:
@jlpadilla did you look at this at all: https://pkg.go.dev/github.com/cloudflare/backoff
Anyway the while the logic is not too complex :
can we measure the effect of how bad it is in scalelab so that we can test how effective the new code is?

jlpadilla:
@joydeep I considered [this similar backoff library.](https://pkg.go.dev/github.com/cenkalti/backoff/v4#pkg-overview)
Because we had a backoff logic, it is less risk to update our existing logic,

I exposed an environment variable to change the jitter, it will help to run experiments with different settings on scale lab.
We have a metric that captures every request, but I don't think we can notice the effect, unless we change the scrape interval to be much smaller.  It will be much simpler to look at the request timestamps in the logs.

SherinV (approved):

jlpadilla:
Tests added for new functions, but threshold not met on legacy code.

PR #261, opened 2023-08-31
Reduce code duplication

jlpadilla:
### Related Issue
N/A

### Description of changes
- Remove duplicate code.


pkg/transforms/transformer.go
```diff
@@ -216,16 +216,8 @@ func TransformRoutine(input chan *Event, output chan NodeEvent) {
 			}
 			trans = CronJobResourceBuilder(&typedResource)
```
jlpadilla, line 218: Combines both `case` into a single with 2 options like `case "a", "b":`

PR #263, opened 2023-09-14
Collect node status field

jlpadilla:
### Related Issue
N/A

### Description of changes
- Added the Node `status` field.

### DIFF
--- a/pkg/transforms/node.go
+++ b/pkg/transforms/node.go
@@ -38,7 +38,7 @@ func NodeResourceBuilder(n *v1.Node) *NodeResource {
 		roles = append(roles, "worker")
 	}
 
-	// sort in alphabetical order to make the ui consistant
+	// sort in alphabetical order to make the ui consistent
 	sort.Strings(roles)
 
 	apiGroupVersion(n.TypeMeta, &node) // add kind, apigroup and version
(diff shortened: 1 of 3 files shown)


pkg/transforms/node.go
```diff
@@ -51,6 +51,14 @@ func NodeResourceBuilder(n *v1.Node) *NodeResource {
 	node.Properties["_systemUUID"] = strings.TrimRight(n.Status.NodeInfo.SystemUUID, "\000")
 	node.Properties["role"] = roles
 
+	conditions := n.Status.Conditions
+	for _, condition := range conditions {
+		if condition.Type == v1.NodeReady {
```
SherinV, line 56: Do we want to capture only the nodeReady status? For a node to be healthy, can others be relevant too - like, KubeletHasSufficientMemory/NoDiskPressure/SufficientPID?

pkg/transforms/node.go
```diff
@@ -51,6 +51,14 @@ func NodeResourceBuilder(n *v1.Node) *NodeResource {
...
+			node.Properties["status"] = string(condition.Status)
```
SherinV, line 57: Here the status will only have True or False. Won't it be better to save it as "Ready"/"NotReady" or "NodeReady"/"NodeNotReady"? Otherwise, simply looking at status (True/False) will not show what it is about.
On the cli, it shows as 'Ready'
image

pkg/transforms/node.go
jlpadilla: I had the same thought, these can be useful, but decided to match the CLI.
Also, it could be too much data to show on the table.
I could change it if you think we should.

pkg/transforms/node.go
jlpadilla: Good catch, I'll change it. The condition can also be "Unknown" and I couldn't find how the CLI displays it, but I'll look again.

pkg/transforms/node.go
SherinV: Maybe we should ask the customer what they mean by node health - do they look at all or one? For now, we can start with this one and add others as needed.

SherinV (approved):

PR #269, opened 2024-01-31
ACM-9746: Reduce excessive logging from Argo apps with missing resources

jlpadilla:
### Related Issue
https://issues.redhat.com/browse/ACM-9746

### Description of changes
- The logic to build edges for ArgoApps was logging excessively when resources aren't found for an app. We have logic to handle this scenario where resources aren't found, but the large volume of warning log messages causes problems.

### DIFF
--- a/pkg/transforms/argoapplication.go
+++ b/pkg/transforms/argoapplication.go
@@ -198,9 +198,6 @@ func (a ArgoApplicationResource) BuildEdges(ns NodeStore) []Edge {
 					})
 				}
 			} else {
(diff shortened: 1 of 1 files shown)


SherinV (approved):
/lgtm
/hold

bjoydeep:
@jlpadilla @SherinV how can we prevent the search collector from going bonkers if a resource has cut loose. If a resource changes every second, can we handle it in some way w/o drowning. What would be the equivalent exponential-backoff analogy?

PR #298, opened 2024-08-16
Add config/operator/cert policy transform

yiraeChristineKim:
Ref: 
Signed-off-by: yiraeChristineKim <yikim@redhat.com>

### Related Issue
https://issues.redhat.com/browse/ACM-13031

### Description of changes
Add properties such as responseAction, compliant, annotations, severity, parentPolicyName, and parentPolicyNs to config/operator/cert policy. 

Common properties:
- managedFields
- annotations
- compliant
- responseAction (Cert won't have this pe)
- severity
- parentPolicyName/parentPolicyNs

OperatorPolicy
- upgradeAvailable
- deploymentAvailable

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,7 @@ require (
 	k8s.io/apimachinery v0.28.2
 	k8s.io/client-go v13.0.0+incompatible
 	k8s.io/helm v2.17.0+incompatible
-	k8s.io/klog/v2 v2.100.1 // indirect
+	k8s.io/klog/v2 v2.100.1
 	k8s.io/utils v0.0.0-20230406110748-d93618cff8a2 // indirect
 	open-cluster-management.io/multicloud-operators-channel v0.8.0
 	open-cluster-management.io/multicloud-operators-subscription v0.8.0 //Use 2.0 when available
--- a/pkg/reconciler/reconciler_test.go
+++ b/pkg/reconciler/reconciler_test.go
@@ -48,7 +48,7 @@ func initTestReconciler() *Reconciler {
 func createNodeEvents() []tr.NodeEvent {
 	events := NodeEdge{}
 	nodeEvents := []tr.NodeEvent{}
-	//First Node
+	// First Node
 	unstructuredInput := unstructured.Unstructured{
 		Object: map[string]interface{}{
 			"kind": "testowner",
@@ -77,7 +77,7 @@ func createNodeEvents() []tr.NodeEvent {
 	events.BuildNode = append(events.BuildNode, podNode)
 	events.BuildEdges = append(events.BuildEdges, podEdges)
 
-	//Convert events to node events
+	// Convert events to node events
 	for i := range events.BuildNode {
 		ne := tr.NodeEvent{
 			Time:         time.Now().Unix(),
@@ -237,10 +237,10 @@ func TestReconcilerRedundant(t *testing.T) {
 
 func TestReconcilerAddEdges(t *testing.T) {
 	testReconciler := initTestReconciler()
-	//Add events
+	// Add events
 	events := createNodeEvents()
 
-	//Input node events to reconciler
+	// Input node events to reconciler
 	go func() {
 		for _, ne := range events {
 			testReconciler.Input <- ne
@@ -250,16 +250,16 @@ func TestReconcilerAddEdges(t *testing.T) {
 	for range events {
 		testReconciler.reconcileNode()
 	}
-	//Build edges
+	// Build edges
 	edgeMap1 := testReconciler.allEdges()
 
-	//Expected edge
+	// Expected edge
 	edgeMap2 := make(map[string]map[string]tr.Edge, 1)
 	edge := tr.Edge{EdgeType: "ownedBy", SourceUID: "local-cluster/5678", DestUID: "local-cluster/1234", SourceKind: "Pod", DestKind: "testowner"}
 	edgeMap2["local-cluster/5678"] = map[string]tr.Edge{}
 	edgeMap2["local-cluster/5678"]["local-cluster/1234"] = edge
 
-	//Check if the actual and expected edges are the same
+	// Check if the actual and expected edges are the same
 	if !reflect.DeepEqual(edgeMap1, edgeMap2) {
 		t.Fatal("Expected edges not found")
 	} else {
@@ -269,17 +269,17 @@ func TestReconcilerAddEdges(t *testing.T) {
 
 func TestReconcilerDiff(t *testing.T) {
 	testReconciler := initTestReconciler()
-	//Add a node to reconciler previous nodes
+	// Add a node to reconciler previous nodes
 	testReconciler.previousNodes["local-cluster/1234"] = tr.Node{
 		UID: "local-cluster/1234",
 		Properties: map[string]interface{}{
 			"very": "important",
 		},
 	}
-	//Add events
+	// Add events
 	events := createNodeEvents()
 
-	//Input node events to reconciler
+	// Input node events to reconciler
 	go func() {
 		for _, ne := range events {
 			testReconciler.Input <- ne
@@ -289,9 +289,9 @@ func TestReconcilerDiff(t *testing.T) {
 	for range events {
 		testReconciler.reconcileNode()
 	}
-	//Compute reconciler diff - this time there should be 1 node and edge to add, 1 node to update
+	// Compute reconciler diff - this time there should be 1 node and edge to add, 1 node to update
 	diff := testReconciler.Diff()
-	//Compute reconciler diff again - this time there shouldn't be any new edges or nodes to add/update
+	// Compute reconciler diff again - this time there shouldn't be any new edges or nodes to add/update
 	nextDiff := testReconciler.Diff()
 
 	if (len(diff.AddNodes) != 1 || len(diff.UpdateNodes) != 1 || len(diff.AddEdges) != 1) ||
@@ -306,7 +306,7 @@ func TestReconcilerComplete(t *testing.T) {
 	input := make(chan *tr.Event)
 	output := make(chan tr.NodeEvent)
 	ts := time.Now().Unix()
-	//Read all files in test-data
+	// Read all files in test-data
 	dir := "../../test-data"
 	files, err := os.ReadDir(dir)
 	if err != nil {
@@ -316,19 +316,19 @@ func TestReconcilerComplete(t *testing.T) {
 	events := make([]tr.Event, 0)
 	var appInput unstructured.Unstructured
 
-	//Variables to keep track of helm release object
+	// Variables to keep track of helm release object
 	var c v1.ConfigMap
 	var rls release.Release
 	rlsFileCount := 0
-	var rlsEvnt = &tr.Event{}
+	rlsEvnt := &tr.Event{}
 
-	//Convert to events
+	// Convert to events
 	for _, file := range files {
 		filePath := dir + "/" + file.Name()
 		if file.Name() != "helmrelease-release.json" {
 			tr.UnmarshalFile(filePath, &appInput, t)
 			appInputLocal := appInput
-			var in = &tr.Event{
+			in := &tr.Event{
 				Time:      ts,
 				Operation: tr.Create,
 				Resource:  &appInputLocal,
@@ -350,7 +350,7 @@ func TestReconcilerComplete(t *testing.T) {
 	testReconciler := initTestReconciler()
 	go tr.TransformRoutine(input, output)
 
-	//Convert events to Node events
+	// Convert events to Node events
 	go func() {
 		for _, ev := range events {
 			localEv := &ev
@@ -378,7 +378,7 @@ func TestReconcilerComplete(t *testing.T) {
 	// Checks the count of nodes and edges based on the JSON files in pkg/test-data
 	// Update counts when the test data is changed
 	// We don't create Nodes for kind = Event
-	const Nodes = 36
+	const Nodes = 39
 	const Edges = 51
 	if len(com.Edges) != Edges || com.TotalEdges != Edges || len(com.Nodes) != Nodes || com.TotalNodes != Nodes {
 		ns := tr.NodeStore{
--- a/pkg/transforms/policy.go
+++ b/pkg/transforms/policy.go
@@ -11,7 +11,11 @@ Copyright (c) 2020 Red Hat, Inc.
 package transforms
 
 import (
-	p "github.com/stolostron/governance-policy-propagator/api/v1"
+	"strings"
+
+	policy "github.com/stolostron/governance-policy-propagator/api/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
 // PolicyResource ...
@@ -20,7 +24,7 @@ type PolicyResource struct {
 }
 
 // PolicyResourceBuilder ...
-func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
+func PolicyResourceBuilder(p *policy.Policy) *PolicyResource {
 	node := transformCommon(p)         // Start off with the common properties
 	apiGroupVersion(p.TypeMeta, &node) // add kind, apigroup and version
 	// Extract the properties specific to this type
@@ -37,6 +41,93 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
 	if okns && okpp {
 		node.Properties["_parentPolicy"] = pnamespace + "/" + ppolicy
 	}
+
+	return &PolicyResource{node: node}
+}
+
+// For cert, config, operator policies.
+// This function returns `annotations`, `_isExternal` for `source`,
+// and `severity`, `compliant`, and `remediationAction`.
+func getPolicyCommonProperties(c *unstructured.Unstructured, node Node) Node {
+	node.Properties["_isExternal"] = false
+
+	for _, m := range c.GetManagedFields() {
+		if m.Manager == "multicluster-operators-subscription" ||
+			strings.Contains(m.Manager, "argocd") {
+			node.Properties["_isExternal"] = true
+
+			break
+		}
+	}
+
+	typeMeta := metav1.TypeMeta{
+		Kind:       c.GetKind(),
+		APIVersion: c.GetAPIVersion(),
+	}
+
+	apiGroupVersion(typeMeta, &node) // add kind, apigroup and version
+
+	node.Properties["compliant"], _, _ = unstructured.NestedString(c.Object, "status", "compliant")
+
+	node.Properties["severity"], _, _ = unstructured.NestedString(c.Object, "spec", "severity")
+
+	node.Properties["remediationAction"], _, _ = unstructured.NestedString(c.Object, "spec", "remediationAction")
+
+	node.Properties["disabled"], _, _ = unstructured.NestedBool(c.Object, "spec", "disabled")
+
+	return node
+}
+
+// For cert, config policies.
+func StandalonePolicyResourceBuilder(c *unstructured.Unstructured) *PolicyResource {
+	node := transformCommon(c) // Start off with the common properties
+
+	return &PolicyResource{node: getPolicyCommonProperties(c, node)}
+}
+
+func OperatorPolicyResourceBuilder(c *unstructured.Unstructured) *PolicyResource {
+	node := transformCommon(c) // Start off with the common properties
+
+	node = getPolicyCommonProperties(c, node)
+
+	var deploymentAvailable bool
+	var upgradeAvailable bool
+
+	conditions, _, _ := unstructured.NestedSlice(c.Object, "status", "conditions")
+	for _, condition := range conditions {
+		mapCondition, ok := condition.(map[string]interface{})
+		if !ok {
+			continue
+		}
+
+		conditionType, found, err := unstructured.NestedString(mapCondition, "type")
+		if !found || err != nil {
+			continue
+		}
+
+		conditionReason, found, err := unstructured.NestedString(mapCondition, "reason")
+		if !found || err != nil {
+			continue
+		}
+
+		if conditionType == "InstallPlanCompliant" {
+			if conditionReason == "InstallPlanRequiresApproval" || conditionReason == "InstallPlanApproved" {
+				upgradeAvailable = true
+			} else {
+				upgradeAvailable = false
+			}
+		} else if conditionType == "DeploymentCompliant" {
+			if conditionReason == "DeploymentsAvailable" {
+				deploymentAvailable = true
+			} else {
+				deploymentAvailable = false
+			}
+		}
+	}
+
+	node.Properties["deploymentAvailable"] = deploymentAvailable
+	node.Properties["upgradeAvailable"] = upgradeAvailable
+
 	return &PolicyResource{node: node}
 }
 
@@ -47,6 +138,6 @@ func (p PolicyResource) BuildNode() Node {
 
 // BuildEdges construct the edges for Policy Resources
 func (p PolicyResource) BuildEdges(ns NodeStore) []Edge {
-	//no op for now to implement interface
+	// no op for now to implement interface
 	return []Edge{}
 }
--- a/pkg/transforms/policy_test.go
+++ b/pkg/transforms/policy_test.go
@@ -14,6 +14,7 @@ import (
 	"testing"
 
 	policy "github.com/stolostron/governance-policy-propagator/api/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
 func TestTransformPolicy(t *testing.T) {
@@ -25,4 +26,65 @@ func TestTransformPolicy(t *testing.T) {
 	AssertEqual("remediationAction", node.Properties["remediationAction"], "enforce", t)
 	AssertEqual("disabled", node.Properties["disabled"], false, t)
 	AssertEqual("numRules", node.Properties["numRules"], 1, t)
+}
+
+func TestTransformConfigPolicy(t *testing.T) {
+	var object map[string]interface{}
+	UnmarshalFile("configurationpolicy.json", &object, t)
+
+	unstructured := &unstructured.Unstructured{
+		Object: object,
+	}
+
+	configResource := StandalonePolicyResourceBuilder(unstructured)
+
+	node := configResource.BuildNode()
+
+	// Test only the fields that exist in policy - the common test will test the other bits
+	AssertEqual("compliant", node.Properties["compliant"], "NonCompliant", t)
+	AssertEqual("remediationAction", node.Properties["remediationAction"], "inform", t)
+	AssertEqual("severity", node.Properties["severity"], "low", t)
+	AssertEqual("disabled", node.Properties["disabled"], false, t)
+	AssertEqual("_isExternal", node.Properties["_isExternal"], true, t)
+}
+
+func TestTransformOperatorPolicy(t *testing.T) {
+	var object map[string]interface{}
+	UnmarshalFile("operatorpolicy.json", &object, t)
+
+	unstructured := &unstructured.Unstructured{
+		Object: object,
+	}
+
+	operatorResource := OperatorPolicyResourceBuilder(unstructured)
+
+	node := operatorResource.BuildNode()
+
+	// Test only the fields that exist in policy - the common test will test the other bits
+	AssertEqual("compliant", node.Properties["compliant"], "NonCompliant", t)
+	AssertEqual("remediationAction", node.Properties["remediationAction"], "inform", t)
+	AssertEqual("severity", node.Properties["severity"], "critical", t)
+	AssertEqual("deploymentAvailable", node.Properties["deploymentAvailable"], false, t)
+	AssertEqual("upgradeAvailable", node.Properties["upgradeAvailable"], true, t)
+	AssertEqual("disabled", node.Properties["disabled"], false, t)
+	AssertEqual("_isExternal", node.Properties["_isExternal"], false, t)
+}
+
+func TestTransformCertPolicy(t *testing.T) {
+	var object map[string]interface{}
+	UnmarshalFile("certpolicy.json", &object, t)
+
+	unstructured := &unstructured.Unstructured{
+		Object: object,
+	}
+
+	certResource := StandalonePolicyResourceBuilder(unstructured)
+
+	node := certResource.BuildNode()
+
+	// Test only the fields that exist in policy - the common test will test the other bits
+	AssertEqual("compliant", node.Properties["compliant"], "Compliant", t)
+	AssertEqual("severity", node.Properties["severity"], "low", t)
+	AssertEqual("disabled", node.Properties["disabled"], false, t)
+	AssertEqual("_isExternal", node.Properties["_isExternal"], true, t)
 }
--- a/pkg/transforms/transformer.go
+++ b/pkg/transforms/transformer.go
@@ -36,7 +36,10 @@ import (
 // Operation is the event operation
 type Operation int
 
-var APPS_OPEN_CLUSTER_MANAGEMENT_IO = "apps.open-cluster-management.io"
+const (
+	APPS_OPEN_CLUSTER_MANAGEMENT_IO   = "apps.open-cluster-management.io"
+	POLICY_OPEN_CLUSTER_MANAGEMENT_IO = "policy.open-cluster-management.io"
+)
 
 const (
 	Create Operation = iota // 0
@@ -131,7 +134,7 @@ type Transformer struct {
 }
 
 var (
-	NonNSResourceMap map[string]struct{} //store non-namespaced resources in this map
+	NonNSResourceMap map[string]struct{} // store non-namespaced resources in this map
 	NonNSResMapMutex = sync.RWMutex{}
 )
 
@@ -151,7 +154,6 @@ func NewTransformer(inputChan chan *Event, outputChan chan NodeEvent, numRoutine
 		Input:  inputChan,
 		Output: outputChan,
 	}
-
 }
 
 // This function processes k8s objects into Nodes, then pass them into the output channel.
@@ -245,7 +247,7 @@ func TransformRoutine(input chan *Event, output chan NodeEvent) {
 			}
 			trans = DeploymentResourceBuilder(&typedResource)
 
-			//This is an ocp specific resource
+			// This is an ocp specific resource
 		case [2]string{"DeploymentConfig", "apps.openshift.io"}:
 			typedResource := ocpapp.DeploymentConfig{}
 			err := runtime.DefaultUnstructuredConverter.
@@ -255,7 +257,7 @@ func TransformRoutine(input chan *Event, output chan NodeEvent) {
 			}
 			trans = DeploymentConfigResourceBuilder(&typedResource)
 
-			//This is the application's HelmCR of kind HelmRelease.
+			// This is the application's HelmCR of kind HelmRelease.
 		case [2]string{"HelmRelease", APPS_OPEN_CLUSTER_MANAGEMENT_IO}:
 			typedResource := appHelmRelease.HelmRelease{}
 			err := runtime.DefaultUnstructuredConverter.
@@ -346,7 +348,7 @@ func TransformRoutine(input chan *Event, output chan NodeEvent) {
 			}
 			trans = PodResourceBuilder(&typedResource)
 
-		case [2]string{"Policy", "policy.open-cluster-management.io"},
+		case [2]string{"Policy", POLICY_OPEN_CLUSTER_MANAGEMENT_IO},
 			[2]string{"Policy", "policies.open-cluster-management.io"}:
 			typedResource := policy.Policy{}
 			err := runtime.DefaultUnstructuredConverter.
@@ -355,6 +357,13 @@ func TransformRoutine(input chan *Event, output chan NodeEvent) {
 				panic(err) // Will be caught by handleRoutineExit
 			}
 			trans = PolicyResourceBuilder(&typedResource)
+
+		case [2]string{"ConfigurationPolicy", POLICY_OPEN_CLUSTER_MANAGEMENT_IO},
+			[2]string{"CertificatePolicy", POLICY_OPEN_CLUSTER_MANAGEMENT_IO}:
+			trans = StandalonePolicyResourceBuilder(event.Resource)
+
+		case [2]string{"OperatorPolicy", POLICY_OPEN_CLUSTER_MANAGEMENT_IO}:
+			trans = OperatorPolicyResourceBuilder(event.Resource)
 
 		case [2]string{"ReplicaSet", "apps"},
 			[2]string{"ReplicaSet", "extensions"}:
--- a/test-data/operatorpolicy.json
+++ b/test-data/operatorpolicy.json
@@ -1,0 +1,226 @@
+{
+  "apiVersion": "policy.open-cluster-management.io/v1beta1",
+  "kind": "OperatorPolicy",
+  "metadata": {
+    "creationTimestamp": "2024-08-15T16:58:38Z",
+    "generation": 1,
+    "annotations": { "apps.open-cluster-management.io/hosting-subscription": "policies/demo-sub" },
+    "labels": {
+      "cluster-name": "local-cluster",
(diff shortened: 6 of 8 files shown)


go.mod
```diff
@@ -3,7 +3,7 @@
 
 module github.com/stolostron/search-collector
 
-go 1.20
```
yiraeChristineKim, line 6: Conflict with `open-cluster-management.io/config-policy-controller`. So, had to

go.mod
```diff
@@ -56,27 +59,28 @@ require (
...
+	sigs.k8s.io/controller-runtime v0.17.5 // indirect
 	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
-	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
-	sigs.k8s.io/yaml v1.3.0 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
+	sigs.k8s.io/yaml v1.4.0 // indirect
 )
 
 replace (
+	github.com/imdario/mergo => dario.cat/mergo v0.3.12
```
yiraeChristineKim, line 81: To resolve conflict

pkg/reconciler/reconciler_test.go
```diff
@@ -378,7 +378,7 @@ func TestReconcilerComplete(t *testing.T) {
 	// Checks the count of nodes and edges based on the JSON files in pkg/test-data
 	// Update counts when the test data is changed
 	// We don't create Nodes for kind = Event
-	const Nodes = 36
+	const Nodes = 39
```
yiraeChristineKim, line 381: 3 more config/operator/cert policies

pkg/transforms/policy.go
```diff
@@ -37,16 +41,129 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
...
+		node.Properties["responseAction"] = remediationAction
+	}
+
+	return node
+}
+
+func ConfigPolicyResourceBuilder(c *unstructured.Unstructured) *PolicyResource {
+	node := transformCommon(c) // Start off with the common properties
+
+	return &PolicyResource{node: getPolicyCommonProperties(c, node)}
```
yiraeChristineKim, line 101: Same as ConfigPolicyResourceBuilder should I merge into one function? or leave this for the future

dhaiducek: I'd personally prefer a single function, maybe `PolicyResourceBuilder`? Then the `switch` statement could optionally have `fallthrough` to hit the common function for Config and Cert.

go.mod
```diff
@@ -22,7 +22,7 @@ require (
 	k8s.io/apimachinery v0.28.2
 	k8s.io/client-go v13.0.0+incompatible
 	k8s.io/helm v2.17.0+incompatible
-	k8s.io/klog/v2 v2.100.1 // indirect
+	k8s.io/klog/v2 v2.100.1
```
dhaiducek, line 25: Do you need to run `go mod tidy`? This doesn't look intentional.

pkg/transforms/policy_test.go
```diff
@@ -14,6 +14,8 @@ import (
 	"testing"
 
 	policy "github.com/stolostron/governance-policy-propagator/api/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
dhaiducek, line 17: nit: I think this would be clearer as `metav1`

pkg/transforms/policy.go
```diff
@@ -11,7 +11,11 @@ Copyright (c) 2020 Red Hat, Inc.
 package transforms
 
 import (
+	"strings"
+
 	p "github.com/stolostron/governance-policy-propagator/api/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
dhaiducek, line 17: nit: I think this would be clearer as `metav1`

pkg/transforms/policy.go
```diff
@@ -11,7 +11,11 @@ Copyright (c) 2020 Red Hat, Inc.
 package transforms
 
 import (
+	"strings"
+
 	p "github.com/stolostron/governance-policy-propagator/api/v1"
```
dhaiducek, line 16: (Optional) This is too short in my opinion--it'd be nice to have `policyv1` (or just `policy`).

pkg/transforms/policy.go
```diff
@@ -37,16 +41,129 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
...
+	}
+
+	severity, found, err := unstructured.NestedString(c.Object, "spec", "severity")
+	if !found || err != nil {
+		node.Properties["severity"] = ""
+	} else {
+		node.Properties["severity"] = severity
+	}
+
+	// responseAction is same as `remediationAction`
```
dhaiducek, line 87: I'm curious: Why the change in name to `responseAction`? Is it predefined or trying to be more generic?

pkg/transforms/policy.go
```diff
@@ -37,16 +41,129 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
...
+				break
+			}
+
+			conditionReason, found, err := unstructured.NestedString(mapCondition, "reason")
+			if !found || err != nil {
+				break
+			}
+
+			if conditionType == "InstallPlanCompliant" {
+				if "InstallPlanRequiresApproval" == conditionReason || conditionReason == "InstallPlanApproved" {
```
dhaiducek, line 131: nit: For consistency
```suggestion
				if conditionReason == "InstallPlanRequiresApproval" || conditionReason == "InstallPlanApproved" {
```

pkg/transforms/policy.go
```diff
@@ -37,16 +41,129 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
...
+					upgradeAvailable = "yes"
+				} else {
+					upgradeAvailable = "no"
+				}
+			}
+
+			if conditionType == "DeploymentCompliant" {
+				if "DeploymentsAvailable" == conditionReason {
```
dhaiducek, line 139: nit: For consistency
```suggestion
				if conditionReason == "DeploymentsAvailable" {
```

pkg/transforms/transformer.go
```diff
@@ -356,6 +355,19 @@ func TransformRoutine(input chan *Event, output chan NodeEvent) {
...
 
+		case [2]string{"ConfigurationPolicy", "policy.open-cluster-management.io"}:
+			trans = ConfigPolicyResourceBuilder(event.Resource)
+
+		case [2]string{"OperatorPolicy", "policy.open-cluster-management.io"}:
+			var err error
+			trans, err = OperatorPolicyResourceBuilder(event.Resource)
+			if err != nil {
+				panic(err) // Will be caught by handleRoutineExit
+			}
```
dhaiducek, line 366: Why is there an error being checked here? I don't see anything other than `nil` being returned in the function.

pkg/transforms/transformer.go
```diff
@@ -356,6 +355,19 @@ func TransformRoutine(input chan *Event, output chan NodeEvent) {
 			}
 			trans = PolicyResourceBuilder(&typedResource)
 
+		case [2]string{"ConfigurationPolicy", "policy.open-cluster-management.io"}:
```
dhaiducek, line 358: (Optional) Looks like there might be a convention to use a variable for the API group similar to `APPS_OPEN_CLUSTER_MANAGEMENT_IO`
```suggestion
		case [2]string{"ConfigurationPolicy", POLICY_OPEN_CLUSTER_MANAGEMENT_IO}:
```

dhaiducek (commented):
Some minor comments, but it looks good! Nice work!

pkg/transforms/policy.go
yiraeChristineKim: `responseAction` includes gatekeeper's `enforcementAction`

mprahl: Let's not use this name for the search collector and use the real name. Response action should just be what is displayed in the new UI.

pkg/transforms/policy.go
```diff
@@ -37,16 +41,129 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
...
+	if found && err == nil {
+		for _, condition := range conditions {
+			mapCondition, ok := condition.(map[string]interface{})
+			if !ok {
+				break
+			}
+
+			conditionType, found, err := unstructured.NestedString(mapCondition, "type")
+			if !found || err != nil {
+				break
```
JustinKuli, line 122: Should these `break`s be `continue`s? I don't think it matters much, the k8s api validation will probably make sure they have these properties, but still, I think it makes more sense to try the next condition rather than stop parsing it completely.

pkg/transforms/policy.go
```diff
@@ -37,16 +41,129 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
...
 	return &PolicyResource{node: node}
 }
 
+// For cert, config, operator policies.
+// This function returns `annotations`, `managedFields` for `source`,
+// and `severity`, `compliant`, `responseAction`, `parentPolicyNs`, `parentPolicyName`
+// If `parentPolicyNs`, `parentPolicyName` do not exist, the policy is a stand-alone policy
+func getPolicyCommonProperties(c *unstructured.Unstructured, node Node) Node {
+	node.Properties["annotations"] = c.GetAnnotations()
+	node.Properties["managedFields"] = c.GetManagedFields()
```
JustinKuli, line 54: Does everything in the Node get indexed or stored? If so, I would rather we slim down what parts of the `managedFields` we want to keep.

go.mod
yiraeChristineKim: This is my go mod tidy result.. I can bring it back

yiraeChristineKim: "k8s.io/klog/v2 should be directgo mod tidy" I am having this warning should I bring it back?

pkg/transforms/policy.go
yiraeChristineKim: Is this what you intend Could you check?

go.mod
dhaiducek: No, it's fine--I was surprised since I didn't see it as an added import, but I guess it just hadn't been run previously.

pkg/transforms/policy.go
JustinKuli: The "isExternal" thing is pretty much what I was thinking. Thanks!

pkg/transforms/policy.go
```diff
@@ -37,6 +41,121 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
 	if okns && okpp {
 		node.Properties["_parentPolicy"] = pnamespace + "/" + ppolicy
 	}
+
+	return &PolicyResource{node: node}
+}
+
+// For cert, config, operator policies.
```
mprahl, line 48: I think this comment needs to be updated based on your latest changes.

pkg/transforms/policy.go
```diff
@@ -37,6 +41,121 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
...
+// This function returns `annotations`, `managedFields` for `source`,
+// and `severity`, `compliant`, `remediationAction`, `parentPolicyNs`, `parentPolicyName`
+// If `parentPolicyNs`, `parentPolicyName` do not exist, the policy is a stand-alone policy
+func getPolicyCommonProperties(c *unstructured.Unstructured, node Node) Node {
+	node.Properties["annotations"] = c.GetAnnotations()
```
mprahl, line 53: To be safe, could you please exclude the `kubectl.kubernetes.io/last-applied-configuration` annotation? That could be very large.

pkg/transforms/policy.go
yiraeChristineKim: getPolicyCommonProperties Also used in Operator too

pkg/transforms/policy.go
```diff
@@ -37,6 +41,121 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
...
+	node.Properties["isExternal"] = false
+
+	for _, m := range c.GetManagedFields() {
+		if m.Manager == "multicluster-operators-subscription" ||
+			strings.Contains(m.Manager, "argocd") {
+			node.Properties["isExternal"] = true
```
mprahl, line 59: In general, I'd rather surface up the data for the console to make the decision because now the logic exists in the console and search collector. I can live with this for now since it does make queries easier from search based on this value.

pkg/transforms/policy.go
```diff
@@ -37,6 +41,121 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
...
+			node.Properties["parentPolicyName"] = stringSlice[1]
+		}
+	}
+
+	complianceState, found, err := unstructured.NestedString(c.Object, "status", "compliant")
+	if !found || err != nil {
+		node.Properties["compliant"] = ""
+	} else {
+		node.Properties["compliant"] = complianceState
+	}
```
mprahl, line 85: Optional but this can be simplified to this since you aren't using the error value and when there is an error it's not found, `complianceState` will be an empty string.
```suggestion
	node.Properties["compliant"], _, _ := unstructured.NestedString(c.Object, "status", "compliant")
```

pkg/transforms/policy.go
```diff
@@ -37,6 +41,121 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
...
+
+	severity, found, err := unstructured.NestedString(c.Object, "spec", "severity")
```
mprahl, line 87: Optional: Same optimization as above

pkg/transforms/policy.go
```diff
@@ -37,6 +41,121 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
...
+	if !found || err != nil {
+		node.Properties["severity"] = ""
+	} else {
+		node.Properties["severity"] = severity
+	}
+
+	// remediationAction is same as `remediationAction`
+	remediationAction, found, err := unstructured.NestedString(c.Object, "spec", "remediationAction")
```
mprahl, line 95: Optional: Same optimization as above

pkg/transforms/policy.go
dhaiducek: @mprahl To your point, with the updated code here the goal (from @JustinKuli's review) was to reduce the fields coming from `managedFields`. Should this instead include a subset of the managed fields and let the console parse it?

pkg/transforms/policy.go
```diff
@@ -37,6 +41,121 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
...
+func OperatorPolicyResourceBuilder(c *unstructured.Unstructured) *PolicyResource {
+	node := transformCommon(c) // Start off with the common properties
+
+	node = getPolicyCommonProperties(c, node)
+
+	var deploymentAvailable string
+	var upgradeAvailable string
+
+	conditions, found, err := unstructured.NestedSlice(c.Object, "status", "conditions")
+	if found && err == nil {
```
mprahl, line 121: Optional: You actually don't need this if statement since if it wasn't found or there was an error, `conditions` is an empty slice so the `for` loop will just not execute.

pkg/transforms/policy.go
```diff
@@ -37,6 +41,121 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
...
+			}
+
+			conditionReason, found, err := unstructured.NestedString(mapCondition, "reason")
+			if !found || err != nil {
+				continue
+			}
+
+			if conditionType == "InstallPlanCompliant" {
+				if conditionReason == "InstallPlanRequiresApproval" || conditionReason == "InstallPlanApproved" {
+					upgradeAvailable = "yes"
```
mprahl, line 140: Could this be a boolean instead of a string?

pkg/transforms/policy.go
```diff
@@ -37,6 +41,121 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
...
+				} else {
+					upgradeAvailable = "no"
+				}
+			}
+
+			if conditionType == "DeploymentCompliant" {
+				if conditionReason == "DeploymentsAvailable" {
+					deploymentAvailable = "yes"
```
mprahl, line 148: Could this be a boolean instead of a string?

pkg/transforms/policy.go
```diff
@@ -37,6 +41,121 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
...
+
+			if conditionType == "InstallPlanCompliant" {
+				if conditionReason == "InstallPlanRequiresApproval" || conditionReason == "InstallPlanApproved" {
+					upgradeAvailable = "yes"
+				} else {
+					upgradeAvailable = "no"
+				}
+			}
+
+			if conditionType == "DeploymentCompliant" {
```
mprahl, line 146: This should be `else if`.

pkg/transforms/policy.go
mprahl: @dhaiducek I think surfacing up the manager names would likely be a good compromise as long as we don't need to filter by if a policy is external through search.

JustinKuli: Yeah, just manager names would be a good compromise. Maybe we can unique-ify that list, if it's not too much trouble. What I'd want to avoid is having 12 entries for the same manager just because it's changed 12 slightly different field... but I'm not really that familiar with the format, only that it seems very noisy in the openshift console when it's not auto-hidden :laughing:

pkg/transforms/policy.go
yiraeChristineKim: any `conditionReason` you have in mind?

mprahl: It's a minor optimization since if the previous if statement is true, this second if statement will always be false.

pkg/transforms/policy.go
yiraeChristineKim: so... with only managers? like string array?

JustinKuli: Yeah, something like `c.GetManagedFields().map(m => m.Manager)`

dhaiducek: @JustinKuli's solution sounds great to me! Or even having manager, operation, subresource, and time (more or less everything but `fieldsV1`, the key that creates the noise/volume).

dhaiducek: I guess part of me is wondering whether, when it's pulled from Search, that the logic in the Console can still read it, so maybe leaving the structure in place might be best. I'm not sure I have the full picture of all the flows in my head though and whether that makes sense.

dhaiducek: (Like maybe I'm thinking too much about it and we should keep it simple like @JustinKuli's solutino 😜 )

pkg/transforms/policy.go
mprahl: Right, but the code isn't collecting all of the managed fields anymore and this documentation says it does.

pkg/transforms/policy.go
```diff
@@ -37,6 +41,107 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
...
+			node.Properties["parentPolicyNs"] = stringSlice[0]
+			node.Properties["parentPolicyName"] = stringSlice[1]
+		}
+	}
+
+	node.Properties["compliant"], _, _ = unstructured.NestedString(c.Object, "status", "compliant")
+
+	node.Properties["severity"], _, _ = unstructured.NestedString(c.Object, "spec", "severity")
+
+	// remediationAction is same as `remediationAction`
```
mprahl, line 89:
```suggestion
```

pkg/transforms/policy.go
yiraeChristineKim: Can we vote? lol because Jutin's way make more simple in frontend

mprahl: @yiraeChristineKim I'm fine for now if you want to have it as "isExternal". I can let others chime in if they disagree.

pkg/transforms/policy.go
```diff
@@ -37,6 +41,106 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
...
+// For cert, config, operator policies.
+// This function returns `annotations`, `managedFields` for `source`,
+// and `severity`, `compliant`, `remediationAction`, `parentPolicyNs`, `parentPolicyName`
+// If `parentPolicyNs`, `parentPolicyName` do not exist, the policy is a stand-alone policy
+func getPolicyCommonProperties(c *unstructured.Unstructured, node Node) Node {
+	annos := c.GetAnnotations()
+	// This annotation is large and useless
+	delete(annos, "kubectl.kubernetes.io/last-applied-configuration")
+
+	node.Properties["annotations"] = annos
```
jlpadilla, line 57: - Annotations are common to all resources, so let's move this to `commonProperties()` in `common.go`
- I like that it excludes the last-applied-configuration, but I want to make it more generic to exclude any other large annotations. Maybe exclude any annotation with more than 64 characters?
- We'll need to update how we query for `annotation`. This will follow the same as `label`.

pkg/transforms/policy.go
yiraeChristineKim: Go isExternal

pkg/transforms/policy.go
yiraeChristineKim: let me change the comment

pkg/transforms/policy.go
jlpadilla: This one is interesting, this is the first case that we have pulling data from managedFields.
Would users understand the meaning of `isExternal`? Without this context, I'd assume it comes from either spec or status.

jlpadilla:
If you want these new properties to show up in the search results in the console, you'll need to update [this file.](https://github.com/stolostron/console/blob/main/frontend/src/routes/Search/searchDefinitions.tsx#L266)
The file defines which columns we want the UI to show for each resource kind.

@mprahl This also applies to your other change to use `additionalPrinterColumns`. Eventually we need to refactor the console to use the same concept of printerColumns, but in the meantime let's add the properties that you need.

pkg/transforms/policy.go
```diff
@@ -37,6 +41,106 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
...
+
+	apiGroupVersion(typeMeta, &node) // add kind, apigroup and version
+
+	policyNsName, ok := c.GetLabels()["policy.open-cluster-management.io/policy"]
+	if ok {
+		stringSlice := strings.Split(policyNsName, ".")
+		// If len(stringSlice) != 2, failed to parse policyNsName
+		if len(stringSlice) == 2 {
+			node.Properties["parentPolicyNs"] = stringSlice[0]
+			node.Properties["parentPolicyName"] = stringSlice[1]
```
jlpadilla, line 81: The parentPolicy should be represented with an edge (or relationship). This is how search "connects" resources.

pkg/transforms/policy.go
```diff
@@ -37,6 +41,106 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
...
+			if conditionReason == "DeploymentsAvailable" {
+				deploymentAvailable = "yes"
+			} else {
+				deploymentAvailable = "no"
+			}
+		}
+	}
+
+	node.Properties["deploymentAvailable"] = deploymentAvailable
+	node.Properties["upgradeAvailable"] = upgradeAvailable
```
jlpadilla, line 142: @mprahl Could the OperatorPolicies crd define these in `additionalPrinterColumns`? Then we won't need this after your other change.

test-data/operatorpolicy.json
```diff
@@ -0,0 +1,225 @@
...
+      {
+        "lastTransitionTime": "2024-08-15T16:58:38Z",
+        "message": "the policy spec is valid",
+        "reason": "PolicyValidated",
+        "status": "True",
+        "type": "ValidPolicySpec"
+      }
+    ],
+    "observedGeneration": 1,
+    "relatedObjects": [
```
jlpadilla, line 129: All these `relatedObjects` could be easily stored and shown by search. This is likely out of scope, but wanted to mention it now that we are looking at it. FYI @mprahl

pkg/transforms/policy.go
mprahl: Good point.

@yiraeChristineKim is this easy information to get in the search query to populated the table?

pkg/transforms/policy.go
mprahl: I think so. I'll look into this later.

test-data/operatorpolicy.json
mprahl: Good point. That would be a nice additional feature.

pkg/transforms/policy.go
yiraeChristineKim: @jlpadilla We don't use this for search-page. In Governance, we are building new tabs for configuration/operator/cert policies

pkg/transforms/policy.go
yiraeChristineKim: @jlpadilla We don't use this for search-page. In Governance, we are building new tabs for configuration/operator/cert policies so This properties just for convenience to search `kind: Policy`

yiraeChristineKim: @jlpadilla Correct me if I am wrong @mprahl  As I know, to get this info we need to use other API. I tried to avoid that. Configuration, operator, and certificate policies already have the parent policy's namespace and name in their labels. I included this to determine whether it's standalone or part of a Policy that contains configuration, operator, or certificate policies as policy-templates.

test-data/operatorpolicy.json
JustinKuli: Just FYI, ConfigurationPolicy and OperatorPolicy use this same schema, but it is different than some OpenShift `relatedObjects`, which use [ObjectReference](https://github.com/openshift/api/blob/2c10e5887729/config/v1/types_cluster_operator.go#L96) (this caused some problems in must-gather for a while)

pkg/transforms/policy.go
mprahl: @yiraeChristineKim could we just leverage the labels in the UI instead since they are already present?

yiraeChristineKim: @mprahl  doable

mprahl: @yiraeChristineKim okay, let's try with that and not index these additional properties.

pkg/transforms/policy.go
jlpadilla: Yes, if you don't expect users to filter by this property from search, then add an underscore to hide it from the search ui.

pkg/transforms/policy.go
jlpadilla: 👍 Good point about not duplicating the information in the label.

FYI, it is the same API to get relationships. You'll just need to update the graphQL query to add `related`.

pkg/transforms/policy.go
```diff
@@ -37,6 +41,98 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
...
 		node.Properties["_parentPolicy"] = pnamespace + "/" + ppolicy
 	}
+
+	return &PolicyResource{node: node}
+}
+
+// For cert, config, operator policies.
+// This function returns `annotations`, `_isExternal` for `source`,
+// and `severity`, `compliant`, `remediationAction`, `parentPolicyNs`, `parentPolicyName`
+// If `parentPolicyNs`, `parentPolicyName` do not exist, the policy is a stand-alone policy
```
mprahl, line 51:
```suggestion
// and `severity`, `compliant`, and `remediationAction`.
```

pkg/transforms/policy.go
```diff
@@ -37,6 +41,98 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
...
+func getPolicyCommonProperties(c *unstructured.Unstructured, node Node) Node {
+	annos := c.GetAnnotations()
+	// This annotation is large and useless
+	delete(annos, "kubectl.kubernetes.io/last-applied-configuration")
```
mprahl, line 55: @yiraeChristineKim I think you missed @jlpadilla 's comment that this should be common to all resources, so this line should go in the `commonProperties` function.

mprahl: Actually, looking at the code, it seems like only the labels are collected for all resources and not the annotations. So I think this is right.

pkg/transforms/policy.go
```diff
@@ -37,6 +41,98 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
...
+		}
+
+		conditionReason, found, err := unstructured.NestedString(mapCondition, "reason")
+		if !found || err != nil {
+			continue
+		}
+
+		if conditionType == "InstallPlanCompliant" {
+			if conditionReason == "InstallPlanRequiresApproval" || conditionReason == "InstallPlanApproved" {
+				upgradeAvailable = "yes"
```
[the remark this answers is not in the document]
yiraeChristineKim: Can you explain the reason?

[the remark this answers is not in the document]
mprahl: @yiraeChristineKim if the condition can only be true or false, then a boolean is the right type. It also makes API interface simpler because you know there can only be two values rather than wonder what the magic values are (e.g. is it `True`, `YES`, `Y`, `yes`, and etc.).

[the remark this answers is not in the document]
yiraeChristineKim: frontend side we can set   ```
deploymentAvailable?: 'yes' | 'no'
  upgradeAvailable?: 'yes' | 'no'
```
so no issue. Bool makes it complicated because I have to use another interface and additional type check. We use **worker** for editing the policies but I am trying to avoid any additional calculations because of performance issues in console.  As far as I know, The Stand-alone table is the only one which uses these properties.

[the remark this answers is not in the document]
Comment (mprahl) [reply pkg/transforms/policy.go:121]: @yiraeChristineKim that shouldn't add much. Let's try with the bool and if it's an issue, we can go back to this approach.

**Comment by mprahl** [comment]
@jlpadilla could you please take another look?

File: pkg/transforms/policy.go [hunk 34]
```diff
@@ -37,6 +41,99 @@ func PolicyResourceBuilder(p *p.Policy) *PolicyResource {
...
+// This function returns `annotations`, `_isExternal` for `source`,
+// and `severity`, `compliant`, and `remediationAction`.
+func getPolicyCommonProperties(c *unstructured.Unstructured, node Node) Node {
+	annos := c.GetAnnotations()
+	// This annotation is large and useless
+	delete(annos, "kubectl.kubernetes.io/last-applied-configuration")
+
+	node.Properties["annotations"] = annos
+
+	node.Properties["_isExternal"] = false
```
jlpadilla, line 58: Would it work if you add this property only when `true`?
It could help to reduce some data. It's only a small impact, but everything counts.

jlpadilla: Well, ignore this comment. I don't think it is worth the additional complexity.
I wouldn't do the same for searchable properties (without the underscore) because users expect to search for `property=false`.

pkg/transforms/policy.go
jlpadilla: @yiraeChristineKim @mprahl I want to resolve these before we merge. We have 2 options, 
- Ideally, I'd like to implement `annotation` to be common to any resource, but this adds the risk of increasing memory.
- Or we could `_annotation` to hide this property and avoid confusing users. My goal is that this will be only temporary while we implement a common `annotation`, so we'll have to update the client code later.

mprahl: @jlpadilla I actually need the annotations for the Gatekeeper constraints too, so if you're good with it, we can make this common. I'd rather not hide it since it could be useful.

jlpadilla: Great! My preference is to make `annotation` common, but it will require some additional work.
Everything else in this PR is good to merge. I suggest that we separate the common annotation to a new PR and merge this one.

mprahl: @yiraeChristineKim can you do something like this?

```diff
diff --git a/pkg/transforms/common.go b/pkg/transforms/common.go
index 57d97c0..64baaca 100644
--- a/pkg/transforms/common.go
+++ b/pkg/transforms/common.go
@@ -40,6 +40,16 @@ func commonProperties(resource v1.Object) map[string]interface{} {
        if resource.GetLabels() != nil {
                ret["label"] = resource.GetLabels()
        }
+
+       annotations := resource.GetAnnotations()
+
+       if annotations != nil {
+               // This annotation is large and useless
+               delete(annotations, "kubectl.kubernetes.io/last-applied-configuration")
+
+               ret["annotations"] = annotations
+       }
+
        if resource.GetNamespace() != "" {
                ret["namespace"] = resource.GetNamespace()
        }
@@ -196,7 +206,8 @@ func edgesByDestinationName(
        destKind string,
        nodeInfo NodeInfo,
        ns NodeStore,
-       seenDests []string) []Edge {
+       seenDests []string,
+) []Edge {
        ret := []Edge{}
        for _, value := range seenDests {
                // Checking against nodeInfo.UID - it gets updated every time edgesByDestinationName is called
@@ -268,7 +279,7 @@ func edgesByDestinationName(
                                        nodeInfo.EdgeType, destKind, nodeInfo.NameSpace+"/"+name)
                        }
                }
-               seenDests = append(seenDests, nodeInfo.UID) //add nodeInfo UID to processed/seen nodes
+               seenDests = append(seenDests, nodeInfo.UID) // add nodeInfo UID to processed/seen nodes
 
                // If the destination node has property _ownerUID, create an edge between the pod and the destination's owner
                // Call the edgesByOwner recursively to create the uses edge
@@ -378,7 +389,6 @@ func edgesByDeployerSubscriber(nodeInfo NodeInfo, ns NodeStore) []Edge {
        }
        ret = findSub(nodeInfo.UID)
        return ret
-
 }
 
 // Build edges from the source node in nodeInfo to all applications/channels in the subscription's metadata.
diff --git a/pkg/transforms/common_test.go b/pkg/transforms/common_test.go
index bf93898..48eed2d 100644
--- a/pkg/transforms/common_test.go
+++ b/pkg/transforms/common_test.go
@@ -18,8 +18,10 @@ import (
        machineryV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-var labels = map[string]string{"app": "test", "fake": "true", "component": "testapp"}
-var timestamp = machineryV1.Now()
+var (
+       labels    = map[string]string{"app": "test", "fake": "true", "component": "testapp"}
+       timestamp = machineryV1.Now()
+)
 
 // Helper function for creating a k8s resource to pass in to tests.
 // In this case it's a pod.
@@ -32,11 +34,11 @@ func CreateGenericResource() machineryV1.Object {
        p.UID = "00aa0000-00aa-00a0-a000-00000a00a0a0"
        p.CreationTimestamp = timestamp
        p.Labels = labels
+       p.Annotations = map[string]string{"hello": "world"}
        return &p
 }
 
 func TestCommonProperties(t *testing.T) {
-
        res := CreateGenericResource()
        timeString := timestamp.UTC().Format(time.RFC3339)
 
@@ -46,6 +48,7 @@ func TestCommonProperties(t *testing.T) {
        AssertEqual("name", cp["name"], interface{}("testpod"), t)
        AssertEqual("namespace", cp["namespace"], interface{}("default"), t)
        AssertEqual("created", cp["created"], interface{}(timeString), t)
+       AssertEqual("annotations", cp["annotations"].(map[string]string)["hello"], "world", t)
 
        noLabels := true
        for key, value := range cp["label"].(map[string]string) {
diff --git a/pkg/transforms/genericResource_test.go b/pkg/transforms/genericResource_test.go
index 9adc9b4..496fb9d 100644
--- a/pkg/transforms/genericResource_test.go
+++ b/pkg/transforms/genericResource_test.go
@@ -5,7 +5,9 @@ package transforms
 import (
        "testing"
 
+       "github.com/stretchr/testify/assert"
        "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+       "k8s.io/apimachinery/pkg/util/sets"
 )
 
 func Test_genericResourceFromConfig(t *testing.T) {
@@ -19,11 +21,30 @@ func Test_genericResourceFromConfig(t *testing.T) {
        AssertEqual("namespace", node.Properties["namespace"], "open-cluster-management", t)
        AssertEqual("created", node.Properties["created"], "2023-08-23T15:54:22Z", t)
 
+       annotations, ok := node.Properties["annotations"].(map[string]string)
+       assert.True(t, ok)
+
+       // Ensure last-applied-configuration is not present
+       expectedAnnotationKeys := sets.New[string](
+               "alm-examples", "capabilities", "categories", "certified", "createdAt", "description", "olm.operatorGroup",
+               "olm.operatorNamespace", "olm.targetNamespaces", "operatorframework.io/initialization-resource",
+               "operatorframework.io/properties", "operatorframework.io/suggested-namespace",
+               "operators.openshift.io/infrastructure-features", "operators.openshift.io/valid-subscription",
+               "operators.operatorframework.io/internal-objects", "support",
+       )
+
+       actualAnnotationKeys := sets.Set[string]{}
+
+       for key := range annotations {
+               actualAnnotationKeys.Insert(key)
+       }
+
+       assert.True(t, expectedAnnotationKeys.Equal(actualAnnotationKeys))
+
        // Verify properties defined in the transform config
        AssertEqual("display", node.Properties["display"], "Advanced Cluster Management for Kubernetes", t)
        AssertEqual("phase", node.Properties["phase"], "Succeeded", t)
        AssertEqual("version", node.Properties["version"], "2.9.0", t)
-
 }
 
 func Test_genericResourceFromConfigVM(t *testing.T) {
@@ -40,5 +61,4 @@ func Test_genericResourceFromConfigVM(t *testing.T) {
        // Verify properties defined in the transform config
        AssertEqual("display", node.Properties["status"], "Running", t)
        AssertEqual("phase", node.Properties["ready"], "True", t)
-
 }
diff --git a/pkg/transforms/genericresource.go b/pkg/transforms/genericresource.go
index c752409..b3e17a4 100644
--- a/pkg/transforms/genericresource.go
+++ b/pkg/transforms/genericresource.go
@@ -99,6 +99,16 @@ func genericProperties(r *unstructured.Unstructured) map[string]interface{} {
        if r.GetLabels() != nil {
                ret["label"] = r.GetLabels()
        }
+
+       annotations := r.GetAnnotations()
+
+       if annotations != nil {
+               // This annotation is large and useless
+               delete(annotations, "kubectl.kubernetes.io/last-applied-configuration")
+
+               ret["annotations"] = annotations
+       }
+
        if r.GetNamespace() != "" {
                ret["namespace"] = r.GetNamespace()
        }
@@ -109,7 +119,6 @@ func genericProperties(r *unstructured.Unstructured) map[string]interface{} {
                ret["_hostingDeployable"] = r.GetAnnotations()["apps.open-cluster-management.io/hosting-deployable"]
        }
        return ret
-
 }
 
 func genericMetadata(r *unstructured.Unstructured) map[string]string {
diff --git a/pkg/transforms/policy.go b/pkg/transforms/policy.go
index c4ddc7b..5ad15f4 100644
--- a/pkg/transforms/policy.go
+++ b/pkg/transforms/policy.go
@@ -49,12 +49,6 @@ func PolicyResourceBuilder(p *policy.Policy) *PolicyResource {
 // This function returns `annotations`, `_isExternal` for `source`,
 // and `severity`, `compliant`, and `remediationAction`.
 func getPolicyCommonProperties(c *unstructured.Unstructured, node Node) Node {
-       annos := c.GetAnnotations()
-       // This annotation is large and useless
-       delete(annos, "kubectl.kubernetes.io/last-applied-configuration")
-
-       node.Properties["annotations"] = annos
-
        node.Properties["_isExternal"] = false
 
        for _, m := range c.GetManagedFields() {
diff --git a/test-data/clusterserviceversion.json b/test-data/clusterserviceversion.json
index 4cc4b2e..c74927d 100644
--- a/test-data/clusterserviceversion.json
+++ b/test-data/clusterserviceversion.json
@@ -18,6 +18,7 @@
             "operators.openshift.io/infrastructure-features": "[\"disconnected\", \"proxy-aware\", \"fips\"]",
             "operators.openshift.io/valid-subscription": "[\"OpenShift Platform Plus\", \"Red Hat Advanced Cluster Management for Kubernetes\"]",
             "operators.operatorframework.io/internal-objects": "[]",
+            "kubectl.kubernetes.io/last-applied-configuration": "something",
             "support": "Red Hat"
         },
         "creationTimestamp": "2023-08-23T15:54:22Z",
```

mprahl: @jlpadilla here is the PR:
https://issues.redhat.com/browse/ACM-13031

mprahl: @jlpadilla here is the PR:
https://github.com/stolostron/search-collector/pull/302

jlpadilla (approved):
/lgtm

PR #299, opened 2024-08-21
Index additional printer columns for Gatekeeper constraints 

mprahl:
The first commit is a slight performance enhancement in the generic informer. It's unrelated to the larger change but I figured I'd include it.

Gatekeeper constraints are CRD based but the CRDs are generated based on
user defined ConstraintTemplate objects. Because of this, indexing the
constraint's `spec.enforcementAction` and `status.totalViolations` is a
bit difficult with the current setup. This makes it so that the CRD's
additionalPrinterColumns are automatically indexed.

The scope of this change is limited to the constraints.gatekeeper.sh API
group, but removing that check in the the `GenericResourceBuilder`
function will enable it for all resources that don't have a custom
transform.

Note that the names from the additionalPrinterColumns in the CRD are
converted to lowerCamelCase style when set as node properties. The
separators are spaces and dashes (e.g. "enforcement action" and
"enforcement-action" become "enforcementAction). Duplicate node
properties are ignored, with a priority on the default generic property
names.

As part of this change, it made sense to also make `syncInformers` event
driven so that it only gets run when there is a likely relevent change
in a CRD (i.e. added, deleted, or the spec changed). This makes the
indexing of new custom resources nearly instant while also sharing the
CRD informer for keeping an up to date mapping of GVRs to additional
printer columns.

Lastly, main.go was refactored a bit to wait for the two main goroutines
to end before exiting. This allows for some clean up when the interrupt
signal is sent to the process. More refactoring to let every goroutine
clean up should be considered.

Relates:
https://issues.redhat.com/browse/ACM-13278
https://issues.redhat.com/browse/ACM-12572

### DIFF
--- a/pkg/informer/runInformers.go
+++ b/pkg/informer/runInformers.go
@@ -3,8 +3,14 @@
 package informer
 
 import (
+	"context"
+	"errors"
+	"fmt"
 	"strings"
+	"sync"
 	"time"
+	"unicode"
+	"unicode/utf8"
 
 	"github.com/golang/glog"
 	"github.com/stolostron/search-collector/pkg/config"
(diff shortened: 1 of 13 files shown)


mprahl:
@jlpadilla this is ready to review, but it's still missing test coverage. Once you agree with how things are implemented/structured, I'll go ahead figure out how I can add the most test coverage possible here.

pkg/informer/runInformers.go
```diff
@@ -13,33 +19,437 @@ import (
...
+			// ExtractProperty.JSONPath field expects them.
+			jsonPath = fmt.Sprintf("{%s}", jsonPath)
+
+			camelCaseName := toLowerCamelCase(name)
+
+			if camelCaseName == "" {
+				continue
+			}
+
+			printerColumns = append(printerColumns, tr.ExtractProperty{Name: camelCaseName, JSONPath: jsonPath})
```
jlpadilla, line 250: Can you add `Type` (string, integer, etc) to ExtractProperty{} ?
Then, we'll need to use the `Type` later when transforming the resource.

pkg/transforms/genericresource.go
```diff
@@ -33,33 +33,57 @@ func GenericResourceBuilder(r *unstructured.Unstructured) *GenericResource {
...
+
+	// Currently, only pull in the additionalPrinterColumns listed in the CRD if it's a Gatekeeper
+	// constraint. Simply removing the `group == "constraints.gatekeeper.sh"` below will enable it
+	// for all other API groups.
+	if !found && group == "constraints.gatekeeper.sh" {
+		transformConfig = ResourceConfig{properties: additionalColumns}
+	}
+
+	for _, prop := range transformConfig.properties {
+		// Skip properties that are already set. This could happen if additionalPrinterColumns
```
jlpadilla, line 45: So, we have a few properties that are a list of stings. It would cause problems if the same property name is used by a different resource as a single string.
It is a short list, of the top of my head some of these properties are 
- container (Pod)
- images (Pod)
- role (Node)
- there's a few more, I'll need to search all transforms for any property defined as string[].

To avoid problems we would need to either:
- ignore properties with this name OR
- wrap the value in a string[]

pkg/informer/runInformers.go
mprahl: @jlpadilla how about changing the `GenericResourceBuilder` assignment to leverage the underlying type found by Go? We know it'll be limited to valid JSON properties since the Kubernetes API returned the object as JSON before it became a `map[string]interface{}` object when unmarshaled?

```go
n.Properties[prop.Name] = fmt.Sprintf("%s", result[0][0])
```

To:

```go
n.Properties[prop.Name] = result[0][0].Interface()
```

pkg/transforms/genericresource.go
mprahl: I think I got them all in the latest push, but let me know if I missed any!

pkg/informer/runInformers.go
mprahl: I ended up going with the approach I mentioned in my previous comment.

mprahl:
@jlpadilla I added unit tests, but it's not particularly practical to add a unit test to `RunInformers` as there is currently no coverage. The `test-on-kind-cluster` tests should have at least verified that `RunInformers` gets run successfully but that doesn't count in the test coverage.

What do you recommend?

mprahl:
@jlpadilla the latest push made this also configurable but I'll have to rebase after #302.

mprahl:
@jlpadilla okay, this is ready for review again!

jlpadilla (approved):
/lgtm
I've tested with the container image and it works great.

Thanks for this awesome contribution! @mprahl

jlpadilla:
/override SonarCould Code Analysis

Overriding quality gate because it's not feasible to meet the coverage requirement for this change.

jlpadilla:
/override "SonarCould Code Analysis"

Overriding quality gate because it's not feasible to meet the coverage requirement for this change.

jlpadilla:


Overriding quality gate because it's not feasible to meet the coverage requirement for this change.

PR #302, opened 2024-08-28
Add annotations to the index

mprahl:
Relates:
https://issues.redhat.com/browse/ACM-13031

### DIFF
--- a/pkg/transforms/common.go
+++ b/pkg/transforms/common.go
@@ -13,10 +13,12 @@ package transforms
 import (
 	"strings"
 	"time"
+	"unicode/utf8"
 
 	"github.com/golang/glog"
 	"github.com/stolostron/search-collector/pkg/config"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	apiTypes "k8s.io/apimachinery/pkg/types"
 )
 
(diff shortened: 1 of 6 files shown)


yiraeChristineKim (approved):

pkg/transforms/common.go
```diff
@@ -40,6 +40,16 @@ func commonProperties(resource v1.Object) map[string]interface{} {
...
 		ret["label"] = resource.GetLabels()
 	}
+
+	annotations := resource.GetAnnotations()
+
+	if annotations != nil {
+		// This annotation is large and useless
+		delete(annotations, "kubectl.kubernetes.io/last-applied-configuration")
+
+		ret["annotations"] = annotations
```
jlpadilla, line 50: Use `annotation` without the (s) to match `label`.

pkg/transforms/common.go
```diff
@@ -40,6 +40,16 @@ func commonProperties(resource v1.Object) map[string]interface{} {
 	if resource.GetLabels() != nil {
 		ret["label"] = resource.GetLabels()
 	}
+
+	annotations := resource.GetAnnotations()
+
+	if annotations != nil {
+		// This annotation is large and useless
+		delete(annotations, "kubectl.kubernetes.io/last-applied-configuration")
```
jlpadilla, line 48: Can we make this more generic to prevent any other annotations with too much data? I'm thinking we could set a max of 64 characters in the value. We could either ignore the annotation completely or truncate the value.

jlpadilla: Querying for annotations with a JSON string in the value will likely cause problems.

jlpadilla:
This is the correct implementation, but I'm worried about increased memory. Here's what we can do.
- Add a temporary gate like we did for the printer columns.
- If you have more capacity I have a plan to reduce the memory. Maybe we can implement something generic that can help with your other projects.

pkg/transforms/common.go
mprahl: Changed in the latest push.

pkg/transforms/common.go
mprahl: I ended up excluding annotations with 64 characters.

mprahl:
> This is the correct implementation, but I'm worried about increased memory. Here's what we can do.
> 
> * Add a temporary gate like we did for the printer columns.
> * If you have more capacity I have a plan to reduce the memory. Maybe we can implement something generic that can help with your other projects.

Do you still want the gate in place with the 64 characters limit?

yiraeChristineKim (approved):

pkg/transforms/common.go
jlpadilla: thanks

pkg/transforms/common.go
jlpadilla: That works for now, we can change it to truncate if we receive complaints about missing annotations.

jlpadilla:
> Do you still want the gate in place with the 64 characters limit?

@mprahl Yes, let's add the gate, there's resources with a lot of annotations, so I want to have more confidence before opening it to all resources.

Let's add an env in config.go to help us enable it for testing in larger environments. What do you think about these?
- COLLECT_ANNOTATIONS
- COLLECT_CRD_PRINTER_COLUMNS

jlpadilla:
Tested the PR image on my dev cluster and overall looks good to merge.

```yaml
apiVersion: search.open-cluster-management.io/v1alpha1
kind: Search
  name: search-v2-operator
  namespace: open-cluster-management
spec:
  dbStorage:
    size: 15Gi
    storageClassName: gp3-csi
  deployments:
    collector:
      imageOverride: >-
        quay.io/stolostron/search-collector:2.12.0-PR302-ae6173eb5ab9f412ec9d4333fa8c08668b178ba9
    database: {}
    indexer: {}
    queryapi: {}
```

Minor problem:
- Annotations with white spaces cause problems with autocomplete. This is a known issue in our UI, but the new annotation makes it more visible.


image

image

mprahl:
@jlpadilla thanks for testing! I made it configurable and disabled by default as requested.

jlpadilla (approved):
/lgtm

Thanks for the contribution and for accommodating my requests!

PR #304, opened 2024-08-30
Use a more generic type for detecting the API group

mprahl:
The previous solution worked for unstructured and default types but not for the Policy kind. This new type works for all.


mprahl:
/cc @jlpadilla 

Sorry about the noise on this one but it's a small one.

zlayne (approved):

PR #307, opened 2024-09-12
Wait for the CRD to be ready before syncing informers

mprahl:
There was a race condition of if the syncInformers function ran before a new CRD was established, an informer for the CRD didn't get created.


zlayne (approved):

jlpadilla:
Thanks for this fix! 
Sorry I missed this, I was OOO Friday.

PR #312, opened 2024-10-03
ACM-14658 Collect desired state of VirtualMachine

jlpadilla:
### Related Issue
https://issues.redhat.com/browse/ACM-14658

### Description of changes
- Collect desired state of Virtual Machine. (spec.running and spec.runStrategy)
- Using the names _specRunning and _specRunStrategy to indicate that this can change. Ideally we should combine these into a single field.

### DIFF
--- a/pkg/transforms/genericResourceConfig.go
+++ b/pkg/transforms/genericResourceConfig.go
@@ -64,17 +64,19 @@ var defaultTransformConfig = map[string]ResourceConfig{
 		properties: []ExtractProperty{
 			{Name: "cpu", JSONPath: `{.spec.template.spec.domain.cpu.cores}`},
 			{Name: "memory", JSONPath: `{.spec.template.spec.domain.memory.guest}`},
-			{Name: "status", JSONPath: `{.status.printableStatus}`},
 			{Name: "ready", JSONPath: `{.status.conditions[?(@.type=='Ready')].status}`},
+			{Name: "status", JSONPath: `{.status.printableStatus}`},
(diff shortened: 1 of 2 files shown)


zlayne (approved):

pkg/transforms/genericResourceConfig.go
```diff
@@ -64,17 +64,19 @@ var defaultTransformConfig = map[string]ResourceConfig{
 		properties: []ExtractProperty{
 			{Name: "cpu", JSONPath: `{.spec.template.spec.domain.cpu.cores}`},
 			{Name: "memory", JSONPath: `{.spec.template.spec.domain.memory.guest}`},
-			{Name: "status", JSONPath: `{.status.printableStatus}`},
 			{Name: "ready", JSONPath: `{.status.conditions[?(@.type=='Ready')].status}`},
+			{Name: "status", JSONPath: `{.status.printableStatus}`},
+			{Name: "_specRunning", JSONPath: `{.spec.running}`},
+			{Name: "_specRunStrategy", JSONPath: `{.spec.runStrategy}`},
```
jlpadilla, line 70: @zlayne What's your opinion on these property names?
- `spec.running` and `spec.runStrategy` are mutually exclusive.
- Ideally I'd like to combine into a property like `desiredState`, but we aren't able to do that using this configuration.
- I'm using the `_` in the names as an indicator that the property name could change in the future.

zlayne: I think that's fine for the time being. This certainly doesn't affect the UI in any way today. And a user should be able to interpret from CLI as needed.

PR #326, opened 2024-10-24
Index Kyverno policies and policy reports

mprahl:
PolicyReport indexing was previously focused on ACM policy results for Insights. This now also accounts for PolicyReport and ClusterPolicyReport objects generated by Kyverno by having the correct numRuleViolations values and edges created between the report and the Kyverno policy.

Relates:
https://issues.redhat.com/browse/ACM-15127


pkg/transforms/common.go
```diff
@@ -248,23 +248,15 @@ func edgesByDestinationName(
...
-				} else if len(destKindInfo) == 1 {
-					name = destKindInfo[0]
-				} else {
-					glog.V(4).Infof("For %s, %s edge not created as %s is not in namespace/name format",
-						nodeInfo.NameSpace+"/"+nodeInfo.Kind+"/"+nodeInfo.Name, nodeInfo.EdgeType, destKind+"/"+name)
-					continue
-				}
+			namespace := nodeInfo.NameSpace
+
+			destKindInfo := strings.SplitN(name, "/", 2)
```
yiraeChristineKim, line 253: Is there any case like only name such as "pod-1" not "pod-1/default"

pkg/transforms/kyverno.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+type KyvernoPolicyResource struct {
+	node Node
+}
+
+func KyvernoPolicyResourceBuilder(p *unstructured.Unstructured) *KyvernoPolicyResource {
+	node := transformCommon(p) // Start off with the common properties
+	gvk := p.GroupVersionKind()
+	node.Properties["kind"] = gvk.Kind
```
yiraeChristineKim, line 16: You can use "`apiGroupVersion(typeMeta, &node) `"

yiraeChristineKim:
Are we gonnar skip cluster violation column? or any idea you have?

mprahl:
> Are we gonna skip cluster violation column? or any idea you have?

@yiraeChristineKim we'll have it. So the flow is to query for the Kyverno policies and get related `PolicyReport` and `ClusterPolicyReport` objects (that's the edges I added). If any of the `PolicyReport` or `ClusterPolicyReport` objects have `numRuleViolations` > 0, then the cluster has a violation for that Kyverno policy. Very similar to Gatekeeper constraints except that the audit violations are in separate objects and potentially in multiple objects (policy reports).

pkg/transforms/policyreport.go
```diff
@@ -51,11 +55,13 @@ func PolicyReportResourceBuilder(pr *PolicyReport) *PolicyReportResource {
...
 	node.Properties["apigroup"] = gvk.Group
 
+	isKyverno := pr.Labels[managedByLabel] == "kyverno"
+
 	// Filter GRC sourced policy violations from node results
 	// Policy details are displayed elsewhere in the UI, displaying them in the PR will result in double counts on pages
 	results := []ReportResults{}
 	for _, result := range pr.Results {
-		if result.Source == "insights" {
+		if result.Source == "insights" || (isKyverno && (result.Result == "fail" || result.Result == "error")) {
```
yiraeChristineKim, line 64: What about `warn`?

mprahl: @yiraeChristineKim I don't think I'd consider it a rule violation if it's a warning, though I haven't seen a warning provided by Kyverno. Even for an audit only policy, it just puts "fail" for a failed audit.

yiraeChristineKim:
> > Are we gonna skip cluster violation column? or any idea you have?
> 
> @yiraeChristineKim we'll have it. So the flow is to query for the Kyverno policies and get related `PolicyReport` and `ClusterPolicyReport` objects (that's the edges I added). If any of the `PolicyReport` or `ClusterPolicyReport` objects have `numRuleViolations` > 0, then the cluster has a violation for that Kyverno policy. Very similar to Gatekeeper constraints except that the audit violations are in separate objects and potentially in multiple objects (policy reports).

Should we fetch both in console?

pkg/transforms/kyverno.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+type KyvernoPolicyResource struct {
+	node Node
+}
+
+func KyvernoPolicyResourceBuilder(p *unstructured.Unstructured) *KyvernoPolicyResource {
```
jlpadilla, line 14: Please add link to the crd for quick reference. 
Is this the correct crd? https://github.com/kyverno/kyverno/blob/main/config/crds/kyverno/kyverno.io_policies.yaml

The crd has several printerColumns. Would it be useful to add those?

mprahl: Thanks! I added the CRD links and added a couple more columns. I don't think the others are needed at the moment.

mprahl:
I'll reopen this PR. I realized that this isn't working on the latest Kyverno version.

PR #329, opened 2024-10-28
Add edges for VAP and VAPB

mprahl:
A ValidationAdmissionPolicyBinding now has an edge to its ValidatingAdmissionPolicy. Additionally, all paramRefs are also added as edges to the ValidationAdmissionPolicyBinding.

Relates:
https://issues.redhat.com/browse/ACM-15249

### DIFF
--- a/pkg/transforms/vapbinding.go
+++ b/pkg/transforms/vapbinding.go
@@ -3,25 +3,32 @@
 package transforms
 
 import (
+	"encoding/json"
 	"strings"
 
+	"github.com/golang/glog"
+	admissionregistration "k8s.io/api/admissionregistration/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 )
+
 type VapBindingResource struct {
-	node Node
+	node     Node
+	paramRef *admissionregistration.ParamRef
 }
 
 // Validating admission policy Binding
 // Ref: https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/
 func VapBindingResourceBuilder(v *unstructured.Unstructured) *VapBindingResource {
-	node := transformCommon(v)         // Start off with the common properties
+	node := transformCommon(v) // Start off with the common properties
 
 	typeMeta := metav1.TypeMeta{
 		Kind:       v.GetKind(),
 		APIVersion: v.GetAPIVersion(),
 	}
-	
+
 	apiGroupVersion(typeMeta, &node) // add kind, apigroup and version
 	// Extract the properties specific to this type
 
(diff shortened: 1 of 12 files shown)


yiraeChristineKim (approved):

pkg/transforms/vap.go
```diff
@@ -0,0 +1,225 @@
...
+
+	typeMeta := metav1.TypeMeta{
+		Kind:       v.GetKind(),
+		APIVersion: v.GetAPIVersion(),
+	}
+
+	apiGroupVersion(typeMeta, &node)
+
+	node.Properties["paramKind_kind"], _, _ = unstructured.NestedString(v.Object, "spec", "paramKind", "kind")
+	node.Properties["paramKind_apiVersion"], _, _ = unstructured.NestedString(v.Object, "spec", "paramKind", "apiVersion")
```
jlpadilla, line 212: If there's no special logic, I prefer to use the genericResourceConfig.
https://github.com/stolostron/search-collector/blob/main/pkg/transforms/genericResourceConfig.go#L32

mprahl:
@jlpadilla could you please review?

pkg/transforms/vapbinding.go
```diff
@@ -51,6 +89,105 @@ func (v VapBindingResource) BuildNode() Node {
...
+	if v.paramRef == nil {
+		return edges
+	}
+
+	policy, ok := ns.ByKindNamespaceName["ValidatingAdmissionPolicy"]["_NONE"][policyName]
+	if !ok {
+		return edges
+	}
+
+	paramKind, ok := policy.Properties["paramKind_kind"].(string)
```
jlpadilla, line 118: The [node object has a Metadata field](https://github.com/stolostron/search-collector/blob/main/pkg/transforms/transformer.go#L60-L65) that should be used to store any data needed to build the edges.
The difference is that 
- anything stored in `node.Properties` is sent to the hub and becomes a searchable field.
- data stored in `node.Metadata` is kept within the collector only, so this is used for things like building edges.

mprahl: Good point. I updated the generic transformer configuration to allow metadata only fields. Let me know what you think!

jlpadilla (approved):
/lgtm

mprahl:
@jlpadilla the PR is rebased. Could you please take another look?

pkg/transforms/vapbinding.go
```diff
@@ -51,6 +89,105 @@ func (v VapBindingResource) BuildNode() Node {
...
+	}
+
+	paramGV, err := schema.ParseGroupVersion(paramApiVersion)
+	if err != nil {
+		return edges
+	}
+
+	namespaceToName := ns.ByKindNamespaceName[paramKind]
+
+	if v.paramRef.Namespace != "" {
```
yiraeChristineKim, line 135: If paramRef.namesapce is empty or '' (clusterscope), 
```
ns.ByKindNamespaceName[paramKind]["_NONE"] need?
```

mprahl: It's just an optimization when the namespace is specified. If it's not specified then all namespaces, including cluster scoped objects in `_NONE` are searched.

jlpadilla (approved):
/lgtm

PR #344, opened 2024-11-13
update konflux run time to once monthly

smcavey:
### Description of changes
- Update konflux runtime to first day of the month for less  konflux pr toil.


smcavey:
search-indexer pr with new konflux cron expression: https://github.com/stolostron/search-indexer/pull/168

PR #349, opened 2024-12-05
Add edge for resources generated by Kyverno

yiraeChristineKim:
https://issues.redhat.com/browse/ACM-15958

### Related Issue

https://issues.redhat.com/browse/ACM-15958

### Description of changes
- Add edge for resources generated by Kyverno


pkg/transforms/common.go
```diff
@@ -156,6 +156,47 @@ func CommonEdges(uid string, ns NodeStore) []Edge {
 
 	// deployer subscriber edges
 	ret = append(ret, edgesByDeployerSubscriber(nodeInfo, ns)...)
+
+	// Kyverno
```
JustinKuli, line 160: I'm wondering if this would be cleaner as a separate function that might return an Edge, similar to the other `edgesBy...` functions.

JustinKuli (approved):
LGTM

dhaiducek (approved):
LGTM!

pkg/transforms/common.go
```diff
@@ -156,6 +156,53 @@ func CommonEdges(uid string, ns NodeStore) []Edge {
...
 	ret = append(ret, edgesByDeployerSubscriber(nodeInfo, ns)...)
+
+	ret = edgesByKyverno(ret, currNode, ns)
+
+	return ret
+}
+
+// Function to create an edge linking any resource with a Kyverno Policy or ClusterPolicy that generates the resource.
+func edgesByKyverno(ret []Edge, currNode Node, ns NodeStore) []Edge {
+	labels, ok := currNode.Properties["label"].(map[string]string)
```
jlpadilla, line 167: This could throw a runtime error if `currNode.Properties["label"]` is nil. I think of a specific scenario where that would happen, but let's add a safety check just in case.

pkg/transforms/policyreport.go
```diff
@@ -117,8 +124,11 @@ func PolicyReportResourceBuilder(pr *PolicyReport) *PolicyReportResource {
 	policyList := policies.UnsortedList()
 	sort.Strings(policyList)
 
-	// "rules" is incorrect since there is a "rule" field in the results, but this is kept for backwards compatibility
-	node.Properties["rules"] = policyList
```
jlpadilla, line 121: This change is equivalent to an API change, so we need to be careful because if a client is using data from `rules` this will cause a break. We haven't made any change like this before, so we'll need to figure out how to handle it.

PolicyReport is used from the console code, but I'm not sure if we use the `rules` field.  The bigger risk is that customers could be using this data.

jlpadilla: FYI @zlayne 
Do you use the `rules` field from PolicyReport in the console?

zlayne: Not the best example becasue there aren't any violations here, but yes we show the rules in the PolicyReport search result table.
Screenshot 2024-12-19 at 8 18 43 AM

yiraeChristineKim: How about change to "ruleNames"

jlpadilla: @yiraeChristineKim Yes, `ruleNames` works, adding a new field won't break any client.

Now I'm also wondering, do we show **incorrect** data in the screenshot that @zlayne added above? If that's the case we can see his as a bug instead of a name change. If the API is not working as intended, then it's easier to justify the breaking change. It's subjective and has a risk, but it's better to return the correct data.

yiraeChristineKim: I check on the console. The rule columns represent policy names which are applied clusterpolicy, policy.  For example `apiVersion: wgpolicyk8s.io/v1beta1
kind: PolicyReport
metadata:
  name: 3d4d8287-57bf-4d79-a5ff-a7cc0774e11b
  namespace: test1
results:
- message: 'validation error: The label `owner` is required. rule require-labels failed
    at path /metadata/labels/owner/'
  policy: require-owner-labels
  result: fail
  rule: require-labels
  scored: true
  severity: medium
  source: kyverno
  timestamp:
    nanos: 0
    seconds: 1734622805`
It is PolicyReport.results.policy not PolicyReport.results.rule

pkg/transforms/common.go
jlpadilla: Thanks for confirming! 

FYI. Interestingly, if I write it differently I get a runtime panic.

```go
props := make(map[string]interface{}, 0)
l, ok := props["label"]
label := l.(map[string]string)
```
```
panic: interface conversion: interface {} is nil, not map[string]string
```

pkg/transforms/policyreport.go
jlpadilla: Ok, so I'm leaning towards treating it as a bug and return the correct data.
Could you create a bug in jira and separate this change to a new PR? I'm sorry for the extra work, but it can help if anyone is affected by this change.

yiraeChristineKim: I created the issue. https://issues.redhat.com/browse/ACM-16775 Could you review this

pkg/transforms/common.go
yiraeChristineKim: `label, ok := l.(map[string]string)`
maybe work?!

jlpadilla: Your code works well, no changes needed here.

My last comment was just FYI to give more context about why I asked to double check this code.

pkg/transforms/policyreport.go
jlpadilla: thank you!

yiraeChristineKim: @jlpadilla Is this good to merge?

pkg/transforms/policyreport.go
```diff
@@ -117,8 +124,11 @@ func PolicyReportResourceBuilder(pr *PolicyReport) *PolicyReportResource {
...
+	ruleList := rules.UnsortedList()
+	sort.Strings(ruleList)
+
+	node.Properties["ruleNames"] = ruleList
+	node.Properties["policies"] = policyList
```
jlpadilla, line 131: 1. Let's keep the name `rules`, based on the discussion in previous comment.
2. Please create a separate branch and PR with this change, everything in this file goes together.

I'm good to merge after these changes.

jlpadilla (changes requested):
Almost ready to merge, I just want to move the bug fix to a separate PR to make it easier if we need to cherry-pick to another release.

pkg/transforms/policyreport.go
jlpadilla: Almost ready, just want to resolve this last comment. https://github.com/stolostron/search-collector/pull/349#discussion_r1894082438

pkg/transforms/policyreport.go
yiraeChristineKim: After this PR merge, I will rebase https://github.com/stolostron/search-collector/pull/365 PR

jlpadilla (approved):
/lgtm

PR #364, opened 2024-12-18
ACM-16246: Store paramRef to recalculate edges

JustinKuli:
Previously, if the paramRef name, namespace, or selector were changed, the edges were not recalculated because no properties on the node were changing.

Refs:
 - https://issues.redhat.com/browse/ACM-16246

### Related Issue

https://issues.redhat.com/browse/ACM-16246

### Description of changes
See above


yiraeChristineKim:
Can you add unit test

JustinKuli:
@yiraeChristineKim I've added a test! It was a bit tricky to find how to trigger this behavior without fully running everything.

pkg/transforms/vapbinding.go
```diff
@@ -80,6 +80,8 @@ func VapBindingResourceBuilder(v *unstructured.Unstructured) *VapBindingResource
 		return binding
 	}
 
+	// A property must be updated, otherwise the reconcile does not update the edges.
+	binding.node.Properties["_paramRef"] = string(paramRefBytes)
```
jlpadilla, line 84: Can you change where we store this data? The `node.Metadata` map is used to store any data that we need later to compute edges, but it won't get synched with the hub. Everything in `node.Properties` will get added to the search database.
```suggestion
	binding.node.Metadata["_paramRef"] = string(paramRefBytes)
```

JustinKuli: I could put it in Metadata, but on its own there it doesn't update the edges. I think I could update https://github.com/stolostron/search-collector/blob/eeab67249590613b4a9399abcd53dc17aef290be/pkg/reconciler/reconciler.go#L337 to also check the metadata, or add this resource type as an exception...

Thanks to the test encouraged by @yiraeChristineKim this is easy to check :laughing:

jlpadilla: Yes, let's update the reconciler code to check the metadata for this change.
I can't recall the exact reason why we don't check changes in metadata, so let's keep the new case as narrow as possible.
```
ne.Node.Properties["kind"] == "VAPBinding" && ne.Node.Metadata["paramRef"] != previousNode.Metadata["paramRef"]
```

JustinKuli: Updated

jlpadilla (approved):
/lgmt

PR #377, opened 2025-01-06
Add gatekeeper mutation edges

yiraeChristineKim:
### Related Issue
https://issues.redhat.com/browse/ACM-16115](https://issues.redhat.com/browse/ACM-16115)

### Description of changes
- Add gatekeeper mutation edges

### DIFF
--- a/pkg/transforms/common.go
+++ b/pkg/transforms/common.go
@@ -101,6 +101,13 @@ func transformCommon(resource v1.Object) Node {
 		Properties: commonProperties(resource),
 		Metadata:   make(map[string]string),
 	}
+
+	// When a resource is mutated by Gatekeeper, add this annotation
+	mutation, ok := resource.GetAnnotations()["gatekeeper.sh/mutations"]
+	if ok {
+		n.Metadata["gatekeeper.sh/mutations"] = mutation
+	}
+
 	n.Metadata["OwnerUID"] = ownerRefUID(resource.GetOwnerReferences())
 	// Adding OwnerReleaseName and Namespace to resources that doesn't have ownerRef but are deployed by a release.
 	if n.Metadata["OwnerUID"] == "" && resource.GetAnnotations()["meta.helm.sh/release-name"] != "" &&
@@ -108,6 +115,7 @@ func transformCommon(resource v1.Object) Node {
 		n.Metadata["OwnerReleaseName"] = resource.GetAnnotations()["meta.helm.sh/release-name"]
 		n.Metadata["OwnerReleaseNamespace"] = resource.GetAnnotations()["meta.helm.sh/release-namespace"]
 	}
+
 	return n
 }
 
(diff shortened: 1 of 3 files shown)


pkg/transforms/common.go
```diff
@@ -206,6 +216,48 @@ func edgesByKyverno(ret []Edge, currNode Node, ns NodeStore) []Edge {
 	return ret
 }
 
+// Function to create an edge linking a resource to Gatekeeper mutations (e.g., Assign, AssignImage) that modify the resource.
+func edgesByGatekeeperMutation(ret []Edge, currNode Node, ns NodeStore) []Edge {
+	annotations, ok := currNode.Properties["annotation"].(map[string]string)
```
jlpadilla, line 221: Can we mode this to `node.Metadata` instead of `node.Properties`?

The difference is that anything added to `node.Properties` is sent to the database and becomes searchable. Data in `node.Metadata` is internal to the collector, so it's better for cases like this one when the data is needed to compute edges or any other processing in the collector.

yiraeChristineKim: Thanks a lot!

pkg/transforms/common.go
```diff
@@ -206,6 +216,43 @@ func edgesByKyverno(ret []Edge, currNode Node, ns NodeStore) []Edge {
...
+	mutationEntries, ok := currNode.Metadata["gatekeeper.sh/mutations"]
+	if !ok || mutationEntries == "" {
+		return ret
+	}
+
+	for _, kindNsName := range strings.Split(mutationEntries, ", ") {
+		parts := strings.Split(kindNsName, "/")
+		if len(parts) != 3 {
+			continue // Skip invalid entries that don't follow the "kind/namespace/name" format
+		}
```
jlpadilla, line 230: I wonder if we should do this validation and the other below BEFORE saving to `node.Metadata`.  If we do it before, processing happens only when the node changes, here it will happen continuously when any node changes.

yiraeChristineKim: gotcha

yiraeChristineKim: If only one annotation is valid, I would like to proceed with the rest. There the code will be duplicated.

jlpadilla: I was thinking you'd only save the valid annotations. So, if only 1 is valid, you'll only have that one in node.Metadata.
Not sure if the invalid annotations are needed for anything else, or if it's likely to have invalid annotations here.

yiraeChristineKim: so do we have to parse that annotation and attach only valid one, and merge, attach as annotation?

jlpadilla: Well, either one works, I was just thinking about efficiency.
If we have to duplicate the `strings.spit()`, then I don't think it's much better to validate before.
Anyways the difference should be very small, so I'm ok keeping it as is.

jlpadilla (approved):
/lgtm

pkg/transforms/common.go
yiraeChristineKim: You always give us good insight Thanks a lot!

PR #400, opened 2025-02-06
ACM-17312 Search Collector Reconcile Metrics

smcavey:
### Related Issue

https://issues.redhat.com/browse/ACM-17312

### Description of changes
- Added metrics server which exposes metrics ResourcesInformedCount and ResourcesSentToIndexerCount.
ResourcesInformedCount captures the count of all informed and transformed resources. ResourcesSentToIndexer captures the count of reconciled resources. The diff between resources we process and discard from local memory as it doesn't carry an update we need to store in the db vs the resources we process and end up in the diff that needs to make its way to the db.

These two metrics, when compared resource_type to resource_type can help identify where the bulk of resource updates are coming from and what portion of those actually make its way out of the collector...i.e. we can identify the resources leading to "wasted" cpu cycles.


smcavey:
Search collector hub service and service monitor: https://github.com/stolostron/search-v2-operator/pull/286

smcavey:
Search Collector hub service and servicemonitor changes: https://github.com/stolostron/search-v2-operator/pull/286

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+var (
+	PromRegistry = prometheus.NewRegistry()
+
+	// ResourcesInformedCount Counts of informed resources before reconciliation
+	ResourcesInformedCount = promauto.With(PromRegistry).NewCounterVec(prometheus.CounterOpts{
+		Name: "search_collector_resources_informed_count",
+		Help: "Total resources received by informers",
+	}, []string{"resource_type"})
```
jlpadilla, line 17: Can you change to `resource_kind` to match the Kubernetes `kind`?

smcavey: So instead of:
`metrics.ResourcesInformedCount.WithLabelValues(ne.Node.ResourceString).Inc()`
we do `metrics.ResourcesInformedCount.WithLabelValues(ne.Node.Properties["kind"]).Inc()` ? Or you want the label to read `kind` instead of `resource_type` ?

smcavey: NVM I gotcha :upside_down_face:

pkg/reconciler/reconciler_test.go
```diff
@@ -91,6 +102,111 @@ func createNodeEvents() []tr.NodeEvent {
...
+		Addr:    config.Cfg.ServerAddress,
+		Handler: router,
+	}
+
+	go func() {
+		if err := srv.ListenAndServe(); err != http.ErrServerClosed {
+			panic(err)
+		}
+	}()
+	time.Sleep(500 * time.Millisecond) // let server start
```
jlpadilla, line 134: - You can use [httptest](https://pkg.go.dev/net/http/httptest) to mock the server. 

- [Here's a sample test from our search-v2-api](https://github.com/stolostron/search-v2-api/blob/main/pkg/server/healthProbes_test.go#L12)

smcavey: mocking the server ended up actually being unnecessary because I was just gathering from the promregistry, not actually executing any GET requests against the metrics endpoint. Did add those health probe tests, though

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,24 @@
...
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	PromRegistry = prometheus.NewRegistry()
+
```
jlpadilla, line 12: We should add another metric that tracks anything received by the informers.
It will be around this function.   https://github.com/stolostron/search-collector/blob/main/pkg/informer/informer.go#L171-L215

jlpadilla: What you have in the reconciler should capture all the events. Maybe the name of the metric is a little misleading.
The overall flow is:
1. **Informer** receives events from kumernetes.
2. **Transform** extracts the data we want to keep. For example labels and pod status.
3. **Reconcile** decides if our state needs to be updated. Example, pod status changed.
4. **Send** syncs the state with the indexer.

smcavey: Yeah I can see it being a little misleading. After tracing the flow figured it would be nice to handle both metrics in the same area, pre and post reconciliation. I could rename `ResourcesSentToIndexerCount` to `ResourcesRenconciledCount`? So we have the metrics `Informed` and `Reconciled` representing all resources and then just those reconciled (to update state).

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,24 @@
...
+	// ResourcesInformedCount Counts of informed resources before reconciliation
+	ResourcesInformedCount = promauto.With(PromRegistry).NewCounterVec(prometheus.CounterOpts{
+		Name: "search_collector_resources_informed_count",
+		Help: "Total resources received by informers",
```
jlpadilla, line 16: Let's make this a little more clear. We could receive 100s of events for a single resource, so we are are counting the events, not the resources.
```suggestion
		Name: "search_collector_events_received_count",
		Help: "Total events received by informers",
```

smcavey: Metric renamed

jlpadilla (approved):
/lgtm

jlpadilla:
/override "Red Hat Konflux / acm-213-enterprise-contract-registry-standard / search-collector-acm-213"
/override "SonarCloud Code Analysis"

smcavey:
2.12 cherry-pick: https://github.com/stolostron/search-collector/pull/422

PR #434, opened 2025-03-06
support of core api group resources in genericresourceconfig transformer

smcavey:
Related story: https://issues.redhat.com/browse/ACM-8620

### Description of changes
- Added support for core api group resources to have resources indexed from genericResource transformer. Namespace moved as an example of support of generic resources.


jlpadilla (approved):
/lgtm

Thanks @smcavey !

PR #451, opened 2025-03-28
ACM-16461 index more resources for kind node

smcavey:
### Related Issue

https://issues.redhat.com/browse/ACM-16461

### Description of changes
- Added more resources to index for kind:Node
- Refactored GenericResourceBuilder to further generalize resource transforming. Permits adding more fields to defaultTransformConfig for resources that have a specific transforms file.

### DIFF
--- a/pkg/transforms/genericResourceConfig.go
+++ b/pkg/transforms/genericResourceConfig.go
@@ -4,6 +4,8 @@ package transforms
 type ExtractProperty struct {
 	Name     string // `json:"name,omitempty"`
 	JSONPath string // `json:"jsonpath,omitempty"`
+	// Denotes this property represents a memory field of a resource and must be converted to bytes for standardization
+	isMemory bool
 	// An internal property to denote this property should be set on the node's metadata instead.
 	metadataOnly bool
 }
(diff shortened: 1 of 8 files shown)


jlpadilla (approved):
/lgtm

pkg/transforms/genericResourceConfig.go
```diff
@@ -4,6 +4,8 @@ package transforms
 type ExtractProperty struct {
 	Name     string // `json:"name,omitempty"`
 	JSONPath string // `json:"jsonpath,omitempty"`
+	// Denotes this property represents a memory field of a resource and must be converted to bytes for standardization
+	isMemory bool
```
jlpadilla, line 8: Let's change to `type` with values string, number, memory

Should we call it `memory` or something more generic like `bytes`?  This could be used for storage too.

smcavey: I think bytes is good. When we get to the ui/api implementation we can document that it's in bytes. Or append a B or Bytes to the api responses if we are returning the base unit to the user in the event they are not requesting any particular unit of measure.

PR #484, opened 2025-03-31
use builtin min func

smcavey:
### Description of changes
- Using builtin min func [introduced in Go 1.21](https://go.dev/blog/go1.21rc) since we bumped Go version.


jlpadilla (approved):
/lgtm

PR #491, opened 2025-04-02
add properties for VMSnapshot resources

zlayne:
### Description of changes
- Added properties for VirtualMachineSnapshot resources
  - ready (boolean)
  - status (string)
  - sourceVM (string)

### DIFF
--- a/pkg/transforms/genericResourceConfig.go
+++ b/pkg/transforms/genericResourceConfig.go
@@ -98,6 +98,15 @@ var defaultTransformConfig = map[string]ResourceConfig{
 			{Name: "vmSize", JSONPath: `{.metadata.labels.\kubevirt\.io/size}`},
 		},
 	},
+	"VirtualMachineSnapshot.snapshot.kubevirt.io": {
+		properties: []ExtractProperty{
+			{Name: "ready", JSONPath: `{.status.conditions[?(@.type=='Ready')].status}`},
+			{Name: "status", JSONPath: `{.status.conditions[?(@.type=='Ready')].reason}`},
+			// {Name: "lastRestored", JSONPath: `{}`}, // From VirtualMachineRestore resource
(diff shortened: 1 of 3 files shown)


pkg/transforms/genericResourceConfig.go
```diff
@@ -98,6 +98,15 @@ var defaultTransformConfig = map[string]ResourceConfig{
 			{Name: "vmSize", JSONPath: `{.metadata.labels.\kubevirt\.io/size}`},
 		},
 	},
+	"VirtualMachineSnapshot.snapshot.kubevirt.io": {
+		properties: []ExtractProperty{
+			{Name: "ready", JSONPath: `{.status.conditions[?(@.type=='Ready')].status}`},
+			{Name: "status", JSONPath: `{.status.conditions[?(@.type=='Ready')].reason}`},
+			// {Name: "lastRestored", JSONPath: `{}`}, // From VirtualMachineRestore resource
```
smcavey, line 105: Is this here as a future TODO for when VirtualMachineRestore is added or something?

smcavey: It looks like transforms/persistentvolume captures an []string with its accessMode attribute, so other string arrays I'd assume should work

smcavey: Oh but yeah I don't see any string arrays defined in the genericResourceConfig so good question how it would behave

zlayne: Yea I wasn't sure how the generic transform would handle it. I see that it comes through as a string separated by ";", which is how labels work I believe. 
ex: `"indications": "noguestagent; online"`

zlayne: I removed the commented out line, and plan to support the indications string in the console to display the array items.

smcavey (approved):

PR #562, opened 2025-04-21
ACM-16129 add clear all bool to request header

smcavey:
### Related Issue

https://issues.redhat.com/browse/ACM-16129

### Description of changes
- Added X-Clear-All header to request to indexer. Depends on https://github.com/stolostron/search-indexer/pull/392 for eliminating the need to first parse request body to determine if request has ClearAll true or ClearAll false.

### DIFF
--- a/pkg/send/sender.go
+++ b/pkg/send/sender.go
@@ -20,6 +20,7 @@ import (
 	"math"
 	"math/big"
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
@@ -163,7 +164,7 @@ func (s *Sender) completePayload() (Payload, int, int) {
 }
 
 // Send will retry after recoverable errors.
-//  - Aggregator busy
+//   - Aggregator busy
 func (s *Sender) sendWithRetry(payload Payload, expectedTotalResources int, expectedTotalEdges int) error {
 	retry := 0
 	for {
@@ -202,7 +203,17 @@ func (s *Sender) send(payload Payload, expectedTotalResources int, expectedTotal
 		return err
 	}
 	payloadBuffer := bytes.NewBuffer(payloadBytes)
-	resp, err := s.httpClient.Post(s.aggregatorURL+s.aggregatorSyncPath, "application/json", payloadBuffer)
+
+	req, err := http.NewRequest("POST", s.aggregatorURL+s.aggregatorSyncPath, payloadBuffer)
+	if err != nil {
+		return err
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("X-Clear-All", strconv.FormatBool(payload.ClearAll))
+	req.Header.Set("X-Request-ID", strconv.Itoa(payload.RequestId))
+
+	resp, err := s.httpClient.Do(req)
 	if resp != nil && resp.Body != nil {
 		// #nosec G307
 		defer resp.Body.Close()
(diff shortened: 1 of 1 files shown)


pkg/send/sender.go
```diff
@@ -202,7 +203,17 @@ func (s *Sender) send(payload Payload, expectedTotalResources int, expectedTotal
...
-	resp, err := s.httpClient.Post(s.aggregatorURL+s.aggregatorSyncPath, "application/json", payloadBuffer)
+
+	req, err := http.NewRequest("POST", s.aggregatorURL+s.aggregatorSyncPath, payloadBuffer)
+	if err != nil {
+		return err
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("X-Clear-All", strconv.FormatBool(payload.ClearAll))
+	req.Header.Set("X-Request-ID", strconv.Itoa(payload.RequestId))
```
jlpadilla, line 214: The `request-id` is meant to help debug and trace the request, but to be honest, I never use it. I'm ok we we remove it.

smcavey: Its gone

pkg/send/sender.go
```diff
@@ -193,16 +180,25 @@ func (s *Sender) sendWithRetry(payload Payload, expectedTotalResources int, expe
...
 	payloadBuffer := bytes.NewBuffer(payloadBytes)
-	resp, err := s.httpClient.Post(s.aggregatorURL+s.aggregatorSyncPath, "application/json", payloadBuffer)
+
+	req, err := http.NewRequest("POST", s.aggregatorURL+s.aggregatorSyncPath, payloadBuffer)
+	if err != nil {
+		return err
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("X-Clear-All", strconv.FormatBool(payload.ClearAll))
```
jlpadilla, line 199: Wondering if there's a better name, maybe `X-Overwrite-State` or `X-Overwrite-Existing`?

The original meaning of `clearAll` was to delete the current state and write new state. We have improved the logic, but the end result is the same, so maybe we can keep it.

smcavey: I like `X-Overwrite-State`. But also `X-Resync-State`, the attribute would then be consistent with the logic in the syncHandler with the request going to either the sync or resync func

jlpadilla: I'm ok with either one.
Maybe `X-Overwrite-State` is more clear in that the state we are sending will overwrite the existing state.

smcavey: I can run with `X-Overwrite-State`

pkg/send/sender.go
```diff
@@ -44,7 +45,6 @@ type Payload struct {
 	AddEdges    []tr.Edge `json:"addEdges,omitempty"`    // List of Edges which must be added
 	DeleteEdges []tr.Edge `json:"deleteEdges,omitempty"` // List of Edges which must be deleted
 	ClearAll    bool      `json:"clearAll,omitempty"`    // Tells the aggregator to clear existing data first.
```
jlpadilla, line 47: So `ClearAll` is still included in the payload. **That's actually good** because it reduces the risk while we update the code in the indexer. Let's try to come back an clean-up after the indexer updates.

jlpadilla (approved):
/lgtm
/hold
@smcavey You can unhold if this is ready to merge.

PR #595, opened 2025-04-29
Add edges for Cert, Config, and Operator Policies

JustinKuli:
These policy kinds report on related resources in their statuses.

Refs:
 - https://issues.redhat.com/browse/ACM-20098

### Related Issue

https://issues.redhat.com/browse/ACM-20098

### Description of changes
- Added edges for Cert, Config, and Operator Policies

### DIFF
--- a/pkg/transforms/policy.go
+++ b/pkg/transforms/policy.go
@@ -11,6 +11,7 @@ Copyright (c) 2020 Red Hat, Inc.
 package transforms
 
 import (
+	"slices"
 	"strings"
 
 	policy "github.com/stolostron/governance-policy-propagator/api/v1"
@@ -80,17 +81,10 @@ func getPolicyCommonProperties(c *unstructured.Unstructured, node Node) Node {
 	return node
 }
 
-// For cert, config policies.
-func StandalonePolicyResourceBuilder(c *unstructured.Unstructured) *PolicyResource {
-	node := transformCommon(c) // Start off with the common properties
-
-	return &PolicyResource{node: getPolicyCommonProperties(c, node)}
-}
-
 func OperatorPolicyResourceBuilder(c *unstructured.Unstructured) *PolicyResource {
 	node := transformCommon(c) // Start off with the common properties
-
 	node = getPolicyCommonProperties(c, node)
+	node = recordRelatedObjects(c, node)
 
 	var deploymentAvailable bool
 	var upgradeAvailable bool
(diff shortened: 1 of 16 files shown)


JustinKuli:
The code coverage check will fail until https://github.com/openshift/release/pull/64390 (or something similar) is merged

pkg/transforms/policy.go
```diff
@@ -130,7 +130,108 @@ func OperatorPolicyResourceBuilder(c *unstructured.Unstructured) *PolicyResource
...
+
+func recordRelatedObjects(c *unstructured.Unstructured, node Node) Node {
+	relatedObjects, found, err := unstructured.NestedSlice(c.Object, "status", "relatedObjects")
+	if !found || err != nil {
+		return node
+	}
+
+	// The node metadata must be a map[string]string, so this puts all related
+	// objects in one entry. This info is *only* to update the metadata in order
+	// to trigger the edges to be re-calculated.
```
jlpadilla, line 157: Is this data only to **trigger** the edge logic? Or is the data here actually used to re-calculate edges?

JustinKuli: Only to trigger the edge logic. The same data is pulled out separately in `BuildEdges`.

It would be possible for `BuildEdges` to parse the metadata string (or update metadata to accept non-strings), if that would be better for some other reason

JustinKuli:
/cc @yiraeChristineKim

just FYI

pkg/transforms/policy.go
```diff
@@ -80,17 +82,15 @@ func getPolicyCommonProperties(c *unstructured.Unstructured, node Node) Node {
...
 }
 
-// For cert, config policies.
-func StandalonePolicyResourceBuilder(c *unstructured.Unstructured) *PolicyResource {
-	node := transformCommon(c) // Start off with the common properties
-
-	return &PolicyResource{node: getPolicyCommonProperties(c, node)}
+type StandalonePolicyResource struct {
+	node   Node
+	policy unstructured.Unstructured
```
jlpadilla, line 87: I think this will keep the entire policy resource in-memory.
The goal of the transforms is to extract only the min data needed so we can release the memory used by the full resource object. Unfortunately we have this pattern in other places, but I don't want to keep adding more.

I'm thinking that the best solution is to change node.Metadata to a `map[string]interface{}` so we aren't limited to only strings. We'll have to see how big is that refactor.

JustinKuli: It didn't end up being too bad of a refactor - adjusting `GetMetadata` made some things more convenient.

pkg/transforms/policy.go
```diff
@@ -23,6 +24,12 @@ type PolicyResource struct {
 	node Node
 }
 
+type policyRelObj struct {
+	kind      string
+	namespace string
+	name      string
+}
```
yiraeChristineKim, line 31: Do we need apiVersion? Some resources have the same kind, so it might be required.

yiraeChristineKim: RelObj is only for search page? or expect to be used in related table in policyDetail page too?

JustinKuli: Kind, Namespace, and Name are all that is required to look something up in the search NodeStore. So if this extracted the apiVersion, it wouldn't be used anywhere. I agree there could be ambiguous cases beyond the scope of this PR (Kyverno and ocm-io both have "Policy", Argo and ACM both have "Application", Hive and ocm-io both have "ClusterClaim").

The intent is to use these related objects in the "Discovered Policy" UI, to view all related resources for the "same" (eg) ConfigurationPolicy across multiple clusters at once. I don't currently intend to re-work the current cluster-specific policy details page to use this.

yiraeChristineKim (approved):

pkg/transforms/policy.go
jlpadilla: @yiraeChristineKim thanks for bringing this up. We should plan to resolve this technical debt💰 before it causes problems, but it's beyond this PR.

pkg/transforms/policy.go
```diff
@@ -23,6 +24,12 @@ type PolicyResource struct {
 	node Node
 }
 
+type policyRelObj struct {
```
jlpadilla, line 27: I'm thinking we should make this more generic to re-use from other transforms. The basic concept is that we are creating a list of edges. We have everything except the UID, which we'll need to resolve later on buildEdges()

- Rename to something like `relatedObjectRef`
- Move to a different file, maybe `commonEdges.go` or `edgeHelper.go`

pkg/transforms/policy.go
```diff
@@ -130,16 +130,172 @@ func OperatorPolicyResourceBuilder(c *unstructured.Unstructured) *PolicyResource
...
-	return []Edge{}
+	kindRaw, ok := p.node.Properties["kind"]
+	if !ok {
+		return []Edge{}
+	}
+
+	policyKind, ok := kindRaw.(string)
+	if !ok {
+		return []Edge{}
+	}
```
jlpadilla, line 269: You can combine these 2 and save a few lines.

```suggestion
	policyKind, ok := p.node.Properties["kind"].(string)
	if !ok {
		return []Edge{}
	}
```

pkg/transforms/policy.go
```diff
@@ -130,16 +130,172 @@ func OperatorPolicyResourceBuilder(c *unstructured.Unstructured) *PolicyResource
...
+		if res, ok := ns.ByKindNamespaceName[obj.kind][namespace][obj.name]; ok {
+			edges = append(edges, Edge{
+				EdgeType:   "relatedResource",
+				SourceKind: policyKind,
+				SourceUID:  p.node.UID,
+				DestKind:   obj.kind,
+				DestUID:    res.UID,
+			})
+		}
+	}
```
jlpadilla, line 298: Please move this to a generic function, so we can reuse it from other transforms. See this comment. https://github.com/stolostron/search-collector/pull/595/files#r2068962456

pkg/transforms/policy.go
jlpadilla: Thanks for the updates!

pkg/transforms/policy.go
JustinKuli: Added type `relatedObject` and func `edgesFromRelatedObjects` to `common.go`.

jlpadilla (approved):
/lgtm
Thank you @JustinKuli !

PR #596, opened 2025-05-01
Add noncompliant resources to policy properties

JustinKuli:
### Related Issue

https://issues.redhat.com/browse/ACM-20098

### Description of changes
- Fixed a crash when comparing related objects in policies
- Adds noncompliant resources to policy properties


JustinKuli:
/cc @dhaiducek @jlpadilla

There is probably a better way to deal with sending the list of noncompliant resources - when the list of labels gets to the console it looks like `cluster-name=local-cluster; cluster-namespace=local-cluster; ...`, but I couldn't see *how*. Do either of you know anything about how that works?

pkg/transforms/policy.go
```diff
@@ -146,51 +146,72 @@ func recordRelatedObjects(c *unstructured.Unstructured, node Node) Node {
...
 
 	slices.SortFunc(objList, func(a, b relatedObject) int {
-		return strings.Compare(a.kind+a.namespace+a.name, b.kind+b.namespace+b.name)
+		return strings.Compare(a.String(), b.String())
 	})
 
+	slices.Sort(nonCompliantList)
+
 	node.Metadata["relObjs"] = objList
+	node.Properties["nonCompliantObjects"] = strings.Join(nonCompliantList, "; ")
```
jlpadilla, line 171: Why storing in node.Properties?
Ideally we should store this data as an edge with `EdgeType = nonCompliant`.  **However, we currently can't query by edgeType, so we'd need to work on that limitation.**

pkg/transforms/common.go
```diff
@@ -548,9 +548,14 @@ func edgesToApplication(nodeInfo NodeInfo, ns NodeStore, UID string, onlyApplica
 }
 
 type relatedObject struct {
-	kind      string
-	namespace string
-	name      string
+	apiVersion string
```
jlpadilla, line 551: Should also add `apiGroup` here.

pkg/transforms/policy.go
jlpadilla: How many nonCompliant objects do you expect to have on a typical policy? Is there a max?
I'm worried about this string getting too big.

pkg/transforms/common.go
```diff
@@ -548,9 +548,14 @@ func edgesToApplication(nodeInfo NodeInfo, ns NodeStore, UID string, onlyApplica
...
+	kind       string
+	namespace  string
+	name       string
+}
+
+func (r relatedObject) String() string {
+	return r.apiVersion + " " + r.kind + " " + r.namespace + " " + r.name
```
jlpadilla, line 558: Should we use a format similar to the kube api for the resource? For example:
```
/namespace/apigroup/apiversion/kind/name
```
Also namespace will be empty for cluster-scoped resources.

pkg/transforms/policy.go
JustinKuli: There's no hard maximum, and it's hard to say what a "typical" policy might have. Because we allow templating, loops, and complicated selectors in our ConfigurationPolicy syntax, it could potentially be a large number. Of course, the idea would be that a non-compliant resource would need to be remediated, so the "normal" state for resources might be compliant.

pkg/transforms/common.go
JustinKuli: This `apiVersion` is both, like in a kubernetes object. Is there a better term for just the `v1` part?

pkg/transforms/common.go
JustinKuli: I'd rather the string format just be for sorting and testing, and have a more structured way to pass the data around, but I need to look into the details on why it wasn't getting to the console the way I was expecting yet.

pkg/transforms/policy.go
JustinKuli: `EdgeType = nonCompliant` would be nice, but it wouldn't help inform users about resources that are desired, but completely missing.

jlpadilla: The `argoapplication.go` has a very similar concept. Please check that transform and let's try to keep it as similar as possible.
In the Argo transform, if the resource exists, we build and edge and for missing resources we store it in `_missingResources`.  [Check the logic here.](https://github.com/stolostron/search-collector/blob/main/pkg/transforms/argoapplication.go#L200-L219)

pkg/transforms/common.go
jlpadilla: You are right, it's ok to keep it combined here. An alternative is to name it `apiGroupVersion`.  If you keep the current name please add a comment like `// format:  apigroup/version` so I remember in the future.

The reason it's slightly confusing is because we divided it so it's easier to search by `apiGroup` or `apiVersion`. [Reference code here.](https://github.com/stolostron/search-collector/blob/main/pkg/transforms/common.go#L596-L605)

pkg/transforms/common.go
jlpadilla: It's possible that the search-api is trying to parse the string like the labels list.  I'd have to debug to confirm.

The uri string can be sorted for testing. But lets check what we do in `argoapplications.go` for consistency.

JustinKuli: I'll work on matching `argoapplications.go` :+1:

pkg/transforms/common.go
```diff
@@ -547,27 +547,45 @@ func edgesToApplication(nodeInfo NodeInfo, ns NodeStore, UID string, onlyApplica
...
 type relatedObject struct {
-	kind      string
-	namespace string
-	name      string
+	Group     string `json:"g,omitempty"`
+	Version   string `json:"v,omitempty"`
+	Kind      string `json:"k,omitempty"`
+	Namespace string `json:"ns,omitempty"`
+	Name      string `json:"n,omitempty"`
+}
```
JustinKuli, line 558: @jlpadilla what do you think of this? The shorter field names save up to 23 characters per object. Maybe it's not worth it.

JustinKuli:
/hold

This was unlikely to be merged immediately, but I need to try some more things with OperatorPolicy.

pkg/transforms/common.go
jlpadilla: It's harder to read, but I like that it makes it harder to incorrectly match the string in a keyword search.

JustinKuli: Oh, that's a really good point! Accidental feature :smile:

jlpadilla: I'm thinking that a base64 encoded string could be even better, but harder to decode. Thoughts?

JustinKuli: base64 has an encoding overhead. Since this is normal text, I'm not sure what we would gain?
```
❯ echo '{"g":"operators.coreos.com","v":"v1alpha1","k":"ClusterServiceVersion","ns":"open-cluster-management","n":"advanced-cluster-management.v2.9.0"}' | wc -c
144

❯ echo '{"g":"operators.coreos.com","v":"v1alpha1","k":"ClusterServiceVersion","ns":"open-cluster-management","n":"advanced-cluster-management.v2.9.0"}' | base64 | wc -c
195
```

jlpadilla: Well, my thought was that base64 would make it harder to get a match by keyword searches, but maybe it isn't worth the overhead. Another con is that it would be harder to explain when we do get an unexpected match.
Let's keep it as is.

pkg/transforms/common.go
```diff
@@ -547,27 +547,45 @@ func edgesToApplication(nodeInfo NodeInfo, ns NodeStore, UID string, onlyApplica
...
+	if r.Group != "" {
+		group = `"g":"` + r.Group + `",`
+	}
+
+	namespace := ""
+	if r.Namespace != "" {
+		namespace = `"ns":"` + r.Namespace + `",`
+	}
+
+	return "{" + group + `"v":"` + r.Version + `","k":"` + r.Kind + `",` + namespace + `"n":"` + r.Name + `"}`
```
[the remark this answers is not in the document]
jlpadilla: Is this still used? I see you are no longer sorting below.

pkg/transforms/policy.go
```diff
@@ -228,19 +258,29 @@ func CertPolicyResourceBuilder(c *unstructured.Unstructured) *PolicyResource {
...
 	})
 
 	node.Metadata["relObjs"] = certList
 
+	nonComp, err := json.Marshal(certList)
+	if err != nil {
+		glog.Error("Policy transform failed to marshal related certpolicy objects", err)
+	}
+
+	node.Properties["_nonCompliantObjects"] = string(nonComp)
```
jlpadilla, line 282: Should this be only nonCompliant objects that are missing?

pkg/transforms/policy.go
```diff
@@ -146,51 +148,79 @@ func recordRelatedObjects(c *unstructured.Unstructured, node Node) Node {
...
+	}
 
 	node.Metadata["relObjs"] = objList
 
+	nonComp, err := json.Marshal(nonCompliantList)
+	if err != nil {
+		glog.Error("Policy transform failed to marshal related objects", err)
+	}
+
+	node.Properties["_nonCompliantObjects"] = string(nonComp)
```
jlpadilla, line 173: Should this be only nonCompliant objects that are missing?

Wondering if you should filter by the reason.
```
"reason": "Resource not found but should exist",
```

jlpadilla: The alternative is to build the list of missing nonCompliant objects while building the edges.

[Here's where we do it for Argo Apps.](https://github.com/stolostron/search-collector/blob/main/pkg/transforms/argoapplication.go#L206-L219)

JustinKuli: I want the table I'm building to show whether the resource is compliant or noncompliant with the policy. So right now I'm not just using this list for the missing objects (but that is one important part of it).

I don't think I can put whether the resource is compliant or not on the resource's own properties in search, because it could be compliant with one policy, and noncompliant with another.

jlpadilla: Here's what I'm thinking.
- if compliant, you should have an edge linking policy and the actual resource.
- if nonCompliant and resource exists, an edge links the policy and actual resource.
- if nonCompliant and resource doesn't exist, add the missing resource in the `_nonCompliant` property.

Ping me and let's talk if it doesn't make sense.

JustinKuli:
@jlpadilla I believe I addressed your latest comments. Can you take another look when you have a chance?

pkg/reconciler/reconciler_test.go
```diff
@@ -473,6 +474,19 @@ func TestReconcilerComplete(t *testing.T) {
...
+		t.Log("Correct number of edges and nodes")
+	}
+
+	// Verify some properties are set during BuildEdges on ConfigurationPolicies
+	configPolNode := ns.ByKindNamespaceName["ConfigurationPolicy"]["local-cluster"]["policy-namespace"]
+
+	missing := configPolNode.Properties["_missingResources"]
+	if missing != `[{"v":"v1","k":"Namespace","n":"nonexistent"}]` {
+		t.Fatal("Incorrect _missingResources; got", missing)
+	}
```
jlpadilla, line 486: This works, or you could also use the assert library.
```suggestion
	assert.Equal(t, configPolNode.Properties["_missingResources"], `[{"v":"v1","k":"Namespace","n":"nonexistent"}]`)
```

jlpadilla (approved):
/lgtm

PR #604, opened 2025-05-06
Add generic transform for VirtualMachineRestore

zlayne:
### Related Issue
https://issues.redhat.com/browse/ACM-0000

### Description of changes
- Collecting more properties for VirtualMachineSnapshot and VirtualMachineRestore resources.

### DIFF
--- a/pkg/transforms/genericResourceConfig.go
+++ b/pkg/transforms/genericResourceConfig.go
@@ -128,7 +128,7 @@ var defaultTransformConfig = map[string]ResourceConfig{
 			{Name: "status", JSONPath: `{.status.conditions[?(@.type=='Ready')].reason}`},
 			{Name: "phase",  JSONPath: `{.status.phase}`},
 			{Name: "indications", JSONPath: `{.status.indications}`}, // this is an array of strings - will collect array items separated by ;
-			{Name: "sourceVM", JSONPath: `{.spec.source.name}`},
+			{Name: "_sourceVM", JSONPath: `{.spec.source.name}`},
 		},
 	},
 	"VirtualMachineRestore.snapshot.kubevirt.io": {
@@ -136,7 +136,7 @@ var defaultTransformConfig = map[string]ResourceConfig{
 			{Name: "ready", JSONPath: `{.status.conditions[?(@.type=='Ready')].status}`},
 			{Name: "status", JSONPath: `{.status.conditions[?(@.type=='Ready')].reason}`},
 			{Name: "lastRestoreTime", JSONPath: `{.status.restoreTime}`},
-			{Name: "targetVM", JSONPath: `{.spec.target.name}`},
+			{Name: "_targetVM", JSONPath: `{.spec.target.name}`},
 		},
 	},
 }
--- a/pkg/transforms/genericResourceConfig.go
+++ b/pkg/transforms/genericResourceConfig.go
@@ -125,18 +125,22 @@ var defaultTransformConfig = map[string]ResourceConfig{
 	"VirtualMachineSnapshot.snapshot.kubevirt.io": {
 		properties: []ExtractProperty{
 			{Name: "ready", JSONPath: `{.status.conditions[?(@.type=='Ready')].status}`},
-			{Name: "status", JSONPath: `{.status.conditions[?(@.type=='Ready')].reason}`},
+			{Name: "_status", JSONPath: `{.status.conditions[?(@.type=='Ready')].reason}`},
 			{Name: "phase",  JSONPath: `{.status.phase}`},
 			{Name: "indications", JSONPath: `{.status.indications}`}, // this is an array of strings - will collect array items separated by ;
-			{Name: "_sourceVM", JSONPath: `{.spec.source.name}`},
+			{Name: "sourceKind", JSONPath: `{.spec.source.kind}`},
+			{Name: "sourceName", JSONPath: `{.spec.source.name}`},
+			{Name: "readyToUse", JSONPath: `{.status.readyToUse}`},
 		},
 	},
 	"VirtualMachineRestore.snapshot.kubevirt.io": {
 		properties: []ExtractProperty{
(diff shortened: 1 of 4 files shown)


pkg/transforms/genericResourceConfig.go
```diff
@@ -126,10 +126,19 @@ var defaultTransformConfig = map[string]ResourceConfig{
 		properties: []ExtractProperty{
 			{Name: "ready", JSONPath: `{.status.conditions[?(@.type=='Ready')].status}`},
 			{Name: "status", JSONPath: `{.status.conditions[?(@.type=='Ready')].reason}`},
+			{Name: "phase",  JSONPath: `{.status.phase}`},
```
smcavey, line 129: Want to add this field to the associated virtualmachinesnapshot test?

zlayne: done

pkg/transforms/genericResourceConfig.go
```diff
@@ -126,10 +126,19 @@ var defaultTransformConfig = map[string]ResourceConfig{
...
 			{Name: "indications", JSONPath: `{.status.indications}`}, // this is an array of strings - will collect array items separated by ;
 			{Name: "sourceVM", JSONPath: `{.spec.source.name}`},
 		},
 	},
+	"VirtualMachineRestore.snapshot.kubevirt.io": {
+		properties: []ExtractProperty{
+			{Name: "ready", JSONPath: `{.status.conditions[?(@.type=='Ready')].status}`},
+			{Name: "status", JSONPath: `{.status.conditions[?(@.type=='Ready')].reason}`},
+			{Name: "lastRestoreTime", JSONPath: `{.status.restoreTime}`},
```
jlpadilla, line 138: This will get collected as a string. I'm not sure if it will work the same as other properties with time. We'd need to test it.

pkg/transforms/genericResourceConfig.go
```diff
@@ -126,10 +126,19 @@ var defaultTransformConfig = map[string]ResourceConfig{
...
+			{Name: "targetVM", JSONPath: `{.spec.target.name}`},
```
jlpadilla, line 139: Would it be better to store this as a relationship?

pkg/transforms/genericResourceConfig.go
zlayne: I have tested in my local env with the PR image. I see the search results with the string as I would expect:
```
                   {
                        ...
                        "kind": "VirtualMachineRestore",
                        "lastRestoreTime": "2025-04-29T18:50:05Z",
                        ...
                    }
```

pkg/transforms/genericResourceConfig.go
zlayne: It should likely be a relationship as well.
However, there are a couple of places in the UI where I need to know the target/source VirtualMachine. With a relationship, I would have to trigger another search request. 
Having it in the resources search data already enables me to bypass having to perform another search query in the UI.

pkg/transforms/genericResourceConfig.go
jlpadilla: Are you able to search like other dates from the UI? Like `lastRestoreTime: <day`

zlayne: Screenshot 2025-05-07 at 2 53 47 PM

pkg/transforms/genericResourceConfig.go
zlayne: It is not currently displayed in the UI. Typically, in VNC, the user will already be in context of the target/source VM when viewing a restore/snapshot resource.

If a user is viewing a restore/snapshot in context of search I think they should either be able to search for target/source VM OR find it via related resource section.

pkg/transforms/genericResourceConfig.go
jlpadilla: Thanks for confirming. Nice expandable UI code! 🥇

pkg/transforms/genericResourceConfig.go
```diff
@@ -126,8 +126,17 @@ var defaultTransformConfig = map[string]ResourceConfig{
...
-			{Name: "sourceVM", JSONPath: `{.spec.source.name}`},
+			{Name: "_sourceVM", JSONPath: `{.spec.source.name}`},
+		},
+	},
+	"VirtualMachineRestore.snapshot.kubevirt.io": {
+		properties: []ExtractProperty{
+			{Name: "ready", JSONPath: `{.status.conditions[?(@.type=='Ready')].status}`},
+			{Name: "status", JSONPath: `{.status.conditions[?(@.type=='Ready')].reason}`},
+			{Name: "lastRestoreTime", JSONPath: `{.status.restoreTime}`},
+			{Name: "_targetVM", JSONPath: `{.spec.target.name}`},
```
jlpadilla, line 139: So, let's match the CLI

1. Rename lastRestoreTime -> restoreTime
2. Rename _targetVM -> targetName
3. Add targetKind   (I wonder if this can be anything other than VirtualMachine)

image

pkg/transforms/genericResourceConfig.go
```diff
@@ -126,8 +126,17 @@ var defaultTransformConfig = map[string]ResourceConfig{
 		properties: []ExtractProperty{
 			{Name: "ready", JSONPath: `{.status.conditions[?(@.type=='Ready')].status}`},
 			{Name: "status", JSONPath: `{.status.conditions[?(@.type=='Ready')].reason}`},
+			{Name: "phase",  JSONPath: `{.status.phase}`},
 			{Name: "indications", JSONPath: `{.status.indications}`}, // this is an array of strings - will collect array items separated by ;
-			{Name: "sourceVM", JSONPath: `{.spec.source.name}`},
+			{Name: "_sourceVM", JSONPath: `{.spec.source.name}`},
```
jlpadilla, line 131: Let's match the CLI here too.

image

jlpadilla (approved):
/lgtm

PR #623, opened 2025-05-20
Fix collection of gatekeeper constraint properties

JustinKuli:
It had been returning too early after not finding the transform config - and the unit test was not actually checking for these properties.

Refs:
 - https://issues.redhat.com/browse/ACM-20932

### Related Issue

https://issues.redhat.com/browse/ACM-20932

### Description of changes
- Fix collection of gatekeeper constraint properties


jlpadilla (approved):
/lgtm

PR #707, opened 2025-06-26
Replace glog with klog

jlpadilla:
### Related Issue
N/A

### Description of changes
- The project now consistently uses klog/v2 for logging instead of the older glog library, which aligns with Kubernetes ecosystem best practices. All logging functionality should work exactly the same, as klog provides the same API as glog but with additional features and better integration with the Kubernetes ecosystem


smcavey:
Ran and appears to still log :)
/lgtm

PR #752, opened 2025-07-25
ACM-22579 easily indexed values of vm, vmi, vmim, pvc

smcavey:
### Related Issue

https://issues.redhat.com/browse/ACM-22579

### Description of changes
- Added indexing of:
pvc: requestedStorage, volumeMode
vm: description, specRunStrategy
vmi: cpu, memory
vmim: statusPhase, endTimestamp

### DIFF
--- a/pkg/transforms/genericResourceConfig.go
+++ b/pkg/transforms/genericResourceConfig.go
@@ -134,7 +134,7 @@ var defaultTransformConfig = map[string]ResourceConfig{
 	},
 	"VirtualMachineInstanceMigration.kubevirt.io": {
 		properties: []ExtractProperty{
-			{Name: "statusPhase", JSONPath: `{.status.phase}`},
+			{Name: "phase", JSONPath: `{.status.phase}`},
 			{Name: "endTime", JSONPath: `{.status.migrationState.endTimestamp}`},
 		},
 	},
--- a/pkg/transforms/genericResourceConfig.go
+++ b/pkg/transforms/genericResourceConfig.go
@@ -74,6 +74,12 @@ var defaultTransformConfig = map[string]ResourceConfig{
 			{Name: "memoryCapacity", JSONPath: `{.status.capacity.memory}`, DataType: DataTypeBytes},
 		},
 	},
+	"PersistentVolumeClaim": {
+		properties: []ExtractProperty{
+			{Name: "requestedStorage", JSONPath: `{.spec.resources.requests.storage}`, DataType: DataTypeBytes},
+			{Name: "volumeMode", JSONPath: `{.spec.volumeMode}`},
+		},
+	},
 	"StorageClass.storage.k8s.io": {
 		properties: []ExtractProperty{
 			{Name: "allowVolumeExpansion", JSONPath: `{.allowVolumeExpansion}`},
@@ -101,10 +107,12 @@ var defaultTransformConfig = map[string]ResourceConfig{
 		properties: []ExtractProperty{
 			{Name: "agentConnected", JSONPath: `{.status.conditions[?(@.type=="AgentConnected")].status}`},
 			{Name: "cpu", JSONPath: `{.spec.template.spec.domain.cpu.cores}`},
+			{Name: "description", JSONPath: `{.metadata.annotations.description}`},
 			{Name: "flavor", JSONPath: `{.spec.template.metadata.annotations.\vm\.kubevirt\.io/flavor}`},
 			{Name: "memory", JSONPath: `{.spec.template.spec.domain.memory.guest}`, DataType: DataTypeBytes},
 			{Name: "osName", JSONPath: `{.spec.template.metadata.annotations.\vm\.kubevirt\.io/os}`},
 			{Name: "ready", JSONPath: `{.status.conditions[?(@.type=='Ready')].status}`},
+			{Name: "specRunStrategy", JSONPath: `{.spec.runStrategy}`},
 			{Name: "status", JSONPath: `{.status.printableStatus}`},
 			{Name: "workload", JSONPath: `{.spec.template.metadata.annotations.\vm\.kubevirt\.io/workload}`},
 			{Name: "_specRunning", JSONPath: `{.spec.running}`},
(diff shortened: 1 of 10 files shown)


pkg/transforms/genericResourceConfig.go
```diff
@@ -101,10 +107,12 @@ var defaultTransformConfig = map[string]ResourceConfig{
 		properties: []ExtractProperty{
 			{Name: "agentConnected", JSONPath: `{.status.conditions[?(@.type=="AgentConnected")].status}`},
 			{Name: "cpu", JSONPath: `{.spec.template.spec.domain.cpu.cores}`},
+			{Name: "description", JSONPath: `{.metadata.annotations.description}`},
 			{Name: "flavor", JSONPath: `{.spec.template.metadata.annotations.\vm\.kubevirt\.io/flavor}`},
 			{Name: "memory", JSONPath: `{.spec.template.spec.domain.memory.guest}`, DataType: DataTypeBytes},
 			{Name: "osName", JSONPath: `{.spec.template.metadata.annotations.\vm\.kubevirt\.io/os}`},
 			{Name: "ready", JSONPath: `{.status.conditions[?(@.type=='Ready')].status}`},
+			{Name: "specRunStrategy", JSONPath: `{.spec.runStrategy}`},
```
jlpadilla, line 115: Maybe this should be just `runStrategy`.

We used `_specRunning` to avoid confusion between the declared and actual state. (spec vs. status).
Then, I think we just tried to be consistent when naming `_specRunStrategy`.

pkg/transforms/genericResourceConfig.go
```diff
@@ -113,20 +121,28 @@ var defaultTransformConfig = map[string]ResourceConfig{
 	},
 	"VirtualMachineInstance.kubevirt.io": {
 		properties: []ExtractProperty{
+			{Name: "cpu", JSONPath: `{.spec.domain.cpu.cores}`},
 			{Name: "ipaddress", JSONPath: `{.status.interfaces[0].ipAddress}`},
 			{Name: "liveMigratable", JSONPath: `{.status.conditions[?(@.type=='LiveMigratable')].status}`},
+			{Name: "memory", JSONPath: `{.spec.domain.memory.guest}`, DataType: DataTypeBytes},
```
jlpadilla, line 127: @zlayne Can you confirm that adding cpu and memory to `VirtualMachineInstance` won't create a problem on the VM List page when you merge it with the VirtualMachine resource, which has the same keys.

I assume both values should always match, but I don't know if it's guaranteed.

pkg/transforms/genericResourceConfig.go
```diff
@@ -113,20 +121,28 @@ var defaultTransformConfig = map[string]ResourceConfig{
...
 			{Name: "osVersion", JSONPath: `{.status.guestOSInfo.version}`},
 			{Name: "phase", JSONPath: `{.status.phase}`},
 			{Name: "ready", JSONPath: `{.status.conditions[?(@.type=='Ready')].status}`},
 			{Name: "vmSize", JSONPath: `{.metadata.labels.\kubevirt\.io/size}`},
 		},
 	},
+	"VirtualMachineInstanceMigration.kubevirt.io": {
+		properties: []ExtractProperty{
+			{Name: "statusPhase", JSONPath: `{.status.phase}`},
+			{Name: "endTimestamp", JSONPath: `{.status.migrationState.endTimestamp}`},
```
jlpadilla, line 138: Should this property be `endTime` ?

We don't have good examples for consistency, but I didn't find any property named `*Timestamp`.
We have 1 property named `restoreTime`.
Other time properties are `created` , `lastShedule`, `updated`, `startedAt`.

smcavey: I can update to `endTime` to add a smidgen of consistency

pkg/transforms/genericResourceConfig.go
zlayne: There won't be any page break issues if CPU & memory are added to the VMI.

Currently, the merged object will have the CPU & memory of whatever resource was parsed last between the VM & VMI. So if we confirm that those values are always the same, then no changes would be necessary. 

If, however, these values can differ between the VM & VMI then I can add some logic in the frontend to only honor the CPU & Memory from the VMI.

pkg/transforms/genericResourceConfig.go
```diff
@@ -101,10 +107,12 @@ var defaultTransformConfig = map[string]ResourceConfig{
 		properties: []ExtractProperty{
 			{Name: "agentConnected", JSONPath: `{.status.conditions[?(@.type=="AgentConnected")].status}`},
 			{Name: "cpu", JSONPath: `{.spec.template.spec.domain.cpu.cores}`},
+			{Name: "description", JSONPath: `{.metadata.annotations.description}`},
```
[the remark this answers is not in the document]
zlayne: Yes it will, I don't think values that contain spaces are commonly searched, but we can create an item to track it.

pkg/transforms/genericResourceConfig.go
```diff
@@ -113,20 +121,28 @@ var defaultTransformConfig = map[string]ResourceConfig{
...
 			{Name: "node", JSONPath: `{.status.nodeName}`},
 			{Name: "osVersion", JSONPath: `{.status.guestOSInfo.version}`},
 			{Name: "phase", JSONPath: `{.status.phase}`},
 			{Name: "ready", JSONPath: `{.status.conditions[?(@.type=='Ready')].status}`},
 			{Name: "vmSize", JSONPath: `{.metadata.labels.\kubevirt\.io/size}`},
 		},
 	},
+	"VirtualMachineInstanceMigration.kubevirt.io": {
+		properties: []ExtractProperty{
+			{Name: "statusPhase", JSONPath: `{.status.phase}`},
```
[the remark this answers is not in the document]
jlpadilla: Right, but I prefer consistency, unless there's a strong reason to deviate. This file has several other cases using `phase`.

pkg/transforms/persistentvolumeclaim_test.go
```diff
@@ -14,17 +14,35 @@ import (
...
-	node := PersistentVolumeClaimResourceBuilder(&p).BuildNode()
+	node := PersistentVolumeClaimResourceBuilder(&p, newUnstructuredPersistentVolumeClaim()).BuildNode()
 
 	// Test only the fields that exist in node - the common test will test the other bits
 	AssertEqual("volumeName", node.Properties["volumeName"], "test-pv", t)
+	AssertEqual("volumeMode", node.Properties["volumeMode"], "Filesystem", t)
 	AssertEqual("status", node.Properties["status"], "Bound", t)
 	AssertEqual("storageClassName", node.Properties["storageClassName"], "test-storage", t)
 	AssertEqual("capacity", node.Properties["capacity"], "5Gi", t)
+	AssertEqual("requestedStorage", node.Properties["requestedStorage"], int64(5368709120), t) // 5 * 1024 * 1024 * 1024
```
jlpadilla, line 31: Same but easier to read.
```suggestion
	AssertEqual("requestedStorage", node.Properties["requestedStorage"], int64(5 * 1024 * 1024 * 1024), t) // 5Gi
```

pkg/transforms/transformer.go
```diff
@@ -328,7 +328,7 @@ func TransformRoutine(input chan *Event, output chan NodeEvent) {
 			if err != nil {
 				panic(err) // Will be caught by handleRoutineExit
 			}
-			trans = PersistentVolumeClaimResourceBuilder(&typedResource)
+			trans = PersistentVolumeClaimResourceBuilder(&typedResource, event.Resource, event.AdditionalPrinterColumns...)
```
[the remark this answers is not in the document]
smcavey: Actually that param in the function is variadic so we can remove that entirely. That should clear up confusion

jlpadilla (approved):
/lgtm
/hold

PR #785, opened 2025-08-25
move dv and pvc from metadata to props on vms

smcavey:
### Description of changes
- Moved dataVolumeNames and pvcClaimNames metadata properties of VirtualMachines onto VirtualMachine properties. Updated configurable edge creation func to be able to build edges also from non-metadata only properties. Discussion: https://redhat-internal.slack.com/archives/C071RR3J7AB/p1756131633904179?thread_ts=1756131020.311809&cid=C071RR3J7AB


pkg/transforms/common.go
```diff
@@ -195,7 +195,12 @@ func edgesByDefaultTransformConfig(ret []Edge, currNode Node, ns NodeStore) []Ed
 
 	if found {
 		for _, e := range transformConfig.edges {
-			val := currNode.Metadata[e.Name]
+			var val interface{}
+			if v, ok := currNode.Metadata[e.Name]; ok {
+				val = v
+			} else if v, ok = currNode.Properties[e.Name]; ok {
+				val = v
+			}
```
jlpadilla, line 203: This works fine for this scenario.

For the future, maybe we'll need to add a field like `FromMetadata` to `ExtractEdge{}` to avoid ambiguity.

PR #839, opened 2026-03-09
ACM-30764 - introduce datatypebool

smcavey:
### Related Issue

https://issues.redhat.com/browse/ACM-30764

### Description of changes
- Added datatype DataTypeBool and store bools as string as the API queries them as if they were strings. We store status condition bools as string, so despite upper vs lowercase when comparing status conditions to these bools, we retain consistency. The only real bool values we retain are _ prefixed and not directly queried via API.

### DIFF
--- a/pkg/transforms/common.go
+++ b/pkg/transforms/common.go
@@ -840,7 +840,15 @@ func applyDefaultTransformConfig(node Node, r *unstructured.Unstructured, additi
 						node.Properties[prop.Name] = interfaceSlice
 					}
 				} else {
-					node.Properties[prop.Name] = val
+					if b, ok := val.(bool); ok { // if value is a bool, store as bool if stated, else as a string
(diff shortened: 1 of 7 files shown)


jlpadilla:
It seems confusing that `DataTypeBool` stores the value as a string.
Maybe a more intuitive behavior is to default to string and convert it to bool only when `DataTypeBool` is specified. Thoughts?

smcavey:
@jlpadilla I like that idea. Pushed changes like that. If value is a bool, unless explicitly flagged as such, we store it as a string. This will permit users to intentionally set values as a bool. But, by doing so, those values can't be queried via the api, that's the core of the issue. So, by letting users do so, they can't query those fields as a bool. I'm inclined to remove the datatypebool and retain storing all bools as string, unless we modify the api to handle querying bools properly, which I don't expect us to do anytime soon.

PR #850, opened 2026-04-08
ACM-30858: collect new Kyverno policy types

jan-law:
Collect 8 new policy types from policies.kyverno.io/v1: validating, mutating, generating and image validating policies along with their namespaced equivalents

These policy types use the same policy reports as ClusterPolicy.

### Related Issue

https://redhat.atlassian.net/browse/ACM-30858

Design doc link in jira issue description

### Description of changes
- Create nodes for new kyverno policies
- Create edges between new policy types, `PolicyReport`s, and objects generated by Kyverno `GeneratingPolicy` and `NamespacedGeneratingPolicy`
- Handle `PolicyReport` node fields when two policy `Kind`s have the same namespaced name
  - Store a new field `_sources`.  `_sources` and `policies` together will uniquely identify the policy that created the report result
  - Prefix `_policyViolationCounts` map keys with source
- When `validationActions` is unset, the node property `validationFailureAction` will be `""`
  - An empty `node.Properties["validationFailureAction"]` either means the kyverno policy type does not support `validationActions`, or the policy will neither Audit or Deny requests during admission. See ValidatingPolicy [CRD](https://github.com/kyverno/kyverno/blob/main/config/crds/policies.kyverno.io/policies.kyverno.io_validatingpolicies.yaml#L565)

## Summary by CodeRabbit

* **New Features**
  * Support for all Kyverno policy kinds (including namespaced variants) and explicit tracking of policy+source pairs for deterministic reporting.

* **Bug Fixes**
  * More accurate detection of Kyverno-generated resources and owner references; broader Kyverno API-group support; improved resolution and sensible defaults for validation/evaluation actions, background, and admission flags.

* **Tests**
  * Expanded unit/integration tests and added a Kyverno PolicyReport test fixture.


jan-law:
@coderabbitai review

jan-law:
@coderabbitai review

jan-law:
@coderabbitai review

pkg/transforms/policyreport.go
```diff
@@ -154,21 +180,47 @@ func (pr PolicyReportResource) BuildEdges(ns NodeStore) []Edge {
...
+			kind = "Policy"
+		case isLegacy && !isNamespaced:
+			kind = "ClusterPolicy"
+		case !isLegacy && isNamespaced:
+			kind = "Namespaced" + policyKind
+		case !isLegacy && !isNamespaced:
+			kind = policyKind
+		default:
+			kind = "ClusterPolicy"
 		}
```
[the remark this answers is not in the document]
jan-law: Not an issue. The edges are built from a report to a Kyverno node whose kind is defined in the switch statement

jan-law:
@coderabbitai review

pkg/transforms/kyverno.go
```diff
@@ -20,36 +22,38 @@ func KyvernoPolicyResourceBuilder(p *unstructured.Unstructured) *KyvernoPolicyRe
...
+			background = bg
+		}
+
+		if admit, ok, _ := unstructured.NestedBool(p.Object, "spec", "admission"); ok {
+			admission = admit
+		}
 	}
 
+	node.Properties["background"] = background
+	node.Properties["admission"] = admission
```
smcavey, line 56: FYI we have a bug open regarding being unable to directly query boolean fields like this in the API https://redhat.atlassian.net/browse/ACM-30764. To query "background: true" or "admission: true", these booleans would have to be stored as strings.

With that bug we are going to wait until 5.0 to change all the existing bools to be stored as strings to resolve.

jan-law: Thanks for letting me know. The main objective of ACM-30858 should not be affected by the bug. We can wait until 5.0

smcavey: If you want you could put their boolean string representations on the node now if you'd like. I think for 5.0 that's what we might do, make a pass of all the booleans and just store them as their string representations with strconv.FormatBool()

pkg/transforms/common.go
jlpadilla: Consider moving this function to `kyverno.go`

pkg/transforms/kyverno.go
jlpadilla, line 56: This is changing the values for background and admission. Wondering if this would be considered a breaking change or a correction of these values?

jan-law: The original values for Kyverno Policy and ClusterPolicy remains the same by reading `spec.background` and `spec.admission`. Default remains true for both properties.

The new policies belong to a new apigroup [policies.kyverno.io](http://policies.kyverno.io/). The equivalent fields are `spec.evaluation.background.enabled` and `spec.evaluation.admission.enabled`

jan-law: Since the other boolean properties will be fixed in 5.0, I’d prefer this PR to keep the same boolean properties for 2.17. I can create a follow up PR for 5.0 if you’d like?

smcavey: That's alright, you can leave them for now and we'll probably handle it come 5.0 with the other stuff, don't worry about it :)

smcavey:
The linter must have updated because we saw new lint errors in https://github.com/stolostron/search-v2-api/pull/707/changes/e6f84e9dfdf2be9c983384d553442d9d95f4c6b4. You could just do something like that to suppress them for now since your changes shouldn't have caused them.

```
diff --git a/pkg/config/config.go b/pkg/config/config.go
index c011cf7..498d27d 100644
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -106,7 +106,7 @@ func InitConfig() {
        setDefaultInt(&Cfg.RetryJitterMS, "RETRY_JITTER_MS", DEFAULT_RETRY_JITTER_MS)
 
        defaultKubePath := filepath.Join(os.Getenv("HOME"), ".kube", "config")
-       if _, err := os.Stat(defaultKubePath); os.IsNotExist(err) {
+       if _, err := os.Stat(defaultKubePath); os.IsNotExist(err) { // #nosec G703
                // set default to empty string if path does not reslove
                defaultKubePath = ""
        }
diff --git a/pkg/informer/runInformers.go b/pkg/informer/runInformers.go
index b09a5e0..9dba1d5 100644
--- a/pkg/informer/runInformers.go
+++ b/pkg/informer/runInformers.go
@@ -598,7 +598,7 @@ func syncInformers(
                        informer.UpdateFunc = createInformerUpdateHandler(gvr)
                        informer.DeleteFunc = informerDeleteHandler
 
-                       informerCtx, informerCancel := context.WithCancel(ctx)
+                       informerCtx, informerCancel := context.WithCancel(ctx) // #nosec G118
                        stoppers[gvr] = informerCancel
                        go informer.Run(informerCtx)
                        // This wait serializes the informer initialization. It is needed to avoid a
```

pkg/transforms/kyverno.go
jlpadilla: thanks for the clarification!

pkg/transforms/common.go
jlpadilla: thank you!

pkg/transforms/transformer.go
```diff
@@ -428,6 +428,18 @@ func TransformRoutine(input chan *Event, output chan NodeEvent) {
...
+
+		case [2]string{"NamespacedImageValidatingPolicy", "policies.kyverno.io"}, [2]string{"ImageValidatingPolicy", "policies.kyverno.io"}:
+			trans = KyvernoPolicyResourceBuilder(event.Resource)
+
+		case [2]string{"NamespacedMutatingPolicy", "policies.kyverno.io"}, [2]string{"MutatingPolicy", "policies.kyverno.io"}:
+			trans = KyvernoPolicyResourceBuilder(event.Resource)
+
+		case [2]string{"NamespacedValidatingPolicy", "policies.kyverno.io"}, [2]string{"ValidatingPolicy", "policies.kyverno.io"}:
+			trans = KyvernoPolicyResourceBuilder(event.Resource)
+
```
jlpadilla, line 438: Slightly simpler.
```suggestion
		case [2]string{"Policy", "kyverno.io"}, [2]string{"ClusterPolicy", "kyverno.io"}:
		case [2]string{"NamespacedGeneratingPolicy", "policies.kyverno.io"}, [2]string{"GeneratingPolicy", "policies.kyverno.io"}:
		case [2]string{"NamespacedImageValidatingPolicy", "policies.kyverno.io"}, [2]string{"ImageValidatingPolicy", "policies.kyverno.io"}:
		case [2]string{"NamespacedMutatingPolicy", "policies.kyverno.io"}, [2]string{"MutatingPolicy", "policies.kyverno.io"}:
		case [2]string{"NamespacedValidatingPolicy", "policies.kyverno.io"}, [2]string{"ValidatingPolicy", "policies.kyverno.io"}:
			trans = KyvernoPolicyResourceBuilder(event.Resource)

```

pkg/transforms/policyreport_test.go
```diff
@@ -18,6 +20,7 @@ func TestTransformPolicyReport(t *testing.T) {
 	// Test unique fields that exist in policy report and are shown in UI - the common test will test the other bits
 	AssertDeepEqual("category Length", len(node.Properties["category"].([]string)), 5, t)
 	AssertDeepEqual("policies", node.Properties["policies"], []string{"policyreport testing risk 1 policy", "policyreport testing risk 2 policy"}, t)
+	AssertDeepEqual("_sources", node.Properties["_sources"], []string{"insights", "insights"}, t)
```
jlpadilla: Would it make sense to remove the duplication? How is this list used? How many items to you expect this list will have on a typical PolicyReport?

jan-law: `_sources` determines the policy `Kind` when building edges; different policy `Kind`s may have the same name. `_sources` will always be the same length as `_policies`.

Originally I did not want to modify `_policies` because I wanted to separate the name and kind. Console uses `_policies` to populate the policy display names. Since the console will need to update the name parsing logic anyway, what do you think of prefixing the policies in `_policies` with the Kind instead?

```
# Before
[_policies] = [“test/my-validating-policy”, “test/my-validating-policy”, “my-cluster-validating-policy”]
[_sources] = [“KyvernoValidatingPolicy”, “KyvernoImageValidatingPolicy”, “KyvernoValidatingPolicy”]

# After
[_policies] = [“NamespacedValidatingPolicy/test/my-validating-policy”, “NamespacedImageValidatingPolicy/test/my-validating-policy”, “ValidatingPolicy/my-cluster-validating-policy”]
```

jlpadilla: Thanks.
The combined `_policies` makes more sense.

JustinKuli (approved):
Looks right to me!

pkg/transforms/policyreport.go
```diff
@@ -108,12 +115,18 @@ func PolicyReportResourceBuilder(pr *PolicyReport) *PolicyReportResource {
...
-		if _, ok := policyViolationCounts[result.Policy]; !ok {
-			policyViolationCounts[result.Policy] = 0
+		// For backwards compatibility, violation keys for Kyverno ClusterPolicy and Policy
+		// are "name" and "namespace/name", respectively
+		if kind == "ClusterPolicy" || kind == "Policy" {
+			policyKey = policyName
+		}
+
+		if _, ok := policyViolationCounts[policyKey]; !ok {
+			policyViolationCounts[policyKey] = 0
```
jan-law, line 125: Note: The console assumes the `policyViolationCounts` keys for ClusterPolicy or Policy match the "name" or "namespace/name" format

```
const policyKey = namespace ? `${namespace}/${name}` : name

for (const violationMapValue of ((policyReport?._policyViolationCounts ?? '') as string).split('; ')) {
  if (!violationMapValue.startsWith(policyKey + '=')) {
    continue
  }
```
https://github.com/stolostron/console/blob/main/frontend/src/routes/Governance/policies/policy-details/PolicyTemplateDetail/PolicyTemplateDetailHooks.tsx#L252

jlpadilla (approved):
/lgtm

PR #865, opened 2026-04-29
ACM-33435: Reduce edge computation - incremental every 5s, full resync every 60s

jlpadilla:
## Summary

- Replaces the previous O(N) `allEdges()` call on every `Diff()` cycle (every 5s) with a hybrid approach
- `incrementalEdges()` is used each cycle — O(|diffNodes|), only recomputing edges for nodes that changed
- `allEdges()` runs periodically (default every 60s) to correct any edges missed by the incremental path (e.g. a Service whose selector now matches a Pod whose labels changed, where the Service itself was not in diffNodes)
- In both cases only the delta (AddEdges/DeleteEdges) is sent to the indexer — no ClearAll / complete payload triggered
- New `EDGE_RESYNC_RATE_MS` env var (default 60 000 ms) controls the full resync interval

## Trade-off

Edges from unchanged nodes that reference a node whose properties changed may take up to `EDGE_RESYNC_RATE_MS` (default 60s) to appear. This is an accepted trade-off: the common case (edges defined on the changed node itself) is still reflected immediately.

## Test plan

- [ ] `make test` passes locally
- [ ] `make lint` passes locally
- [ ] Verify incremental path is used between resync intervals (klog V4)
- [ ] Verify full resync fires every ~60s and sends any corrected edges as a diff (not a complete payload)

🤖 I'm an AI agent. Generated with [Claude Code](https://claude.ai/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added EDGE_RESYNC_RATE_MS configuration (default 60s) to control how often a full edge resynchronization runs.
  * Edge processing now uses a hybrid strategy: scheduled full resynchronizations at the configured interval plus incremental updates for changed nodes between full cycles, reducing unnecessary full recalculations and improving update efficiency.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -32,12 +32,13 @@ const (
 	DEFAULT_AGGREGATOR_PORT    = "3010"
 	DEFAULT_CLUSTER_NAME       = "local-cluster"
 	DEFAULT_POD_NAMESPACE      = "open-cluster-management"
-	DEFAULT_HEARTBEAT_MS       = 300000 // 5 min
-	DEFAULT_MAX_BACKOFF_MS     = 600000 // 10 min
-	DEFAULT_REDISCOVER_RATE_MS = 120000 // 2 min
-	DEFAULT_REPORT_RATE_MS     = 5000   // 5 seconds
-	DEFAULT_RETRY_JITTER_MS    = 5000   // 5 seconds
-	DEFAULT_RUNTIME_MODE       = "production"
+	DEFAULT_HEARTBEAT_MS        = 300000 // 5 min
+	DEFAULT_MAX_BACKOFF_MS      = 600000 // 10 min
+	DEFAULT_REDISCOVER_RATE_MS  = 120000 // 2 min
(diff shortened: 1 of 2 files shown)


jlpadilla:
🤖 I'm an AI agent. Addressed both CodeRabbit review comments:

**1. Copy-on-write edge buckets (O(|currentEdges|) → O(|diffNodes|))**

Replaced the full deep-copy of \`previousEdges\` with a copy-on-write approach. \`incrementalEdges()\` now starts with shallow references to all existing buckets and only deep-copies a bucket on its first write. Unmodified buckets are never allocated.

This required making the edge comparison in \`Diff()\` non-destructive — the old loops deleted matched entries from \`r.previousEdges\` in-place, which would corrupt shared bucket references. The rewritten loops check membership without mutation and also resolve the long-standing TODO about combining the two loops.

**2. Application-first ordering in \`incrementalEdges()\`**

Applied the same Application-first ordering that \`allEdges()\` uses. Diffed nodes are now split into Application UIDs and other UIDs, with Applications processed first. This ensures \`_hostingApplication\` metadata is populated before Subscription edge functions run when both kinds appear in the same diff cycle.

PR #866, opened 2026-05-04
ACM-21531 -  CollectorConfig NamespaceSelector

smcavey:
### Related Issue

https://issues.redhat.com/browse/ACM-21531

### Description of changes
- Parses informer events, filters namespace from CollectorConfig's NamespaceSelector.

## Summary by CodeRabbit

* **New Features**
  * Namespace allowlisting for resource collection, resolved at runtime with a TTL-based cache; falls back to “allow all” if collection is disabled or config errors occur.
  * Two-stage filtering: selectors (labels/expressions) then include/exclude glob patterns; cluster-scoped resources always allowed.

* **Bug Fixes**
  * Initial cache now keeps only allowed namespaces and removes previously-cached disallowed resources.
  * Runtime processing skips ADDED/MODIFIED/DELETED events for disallowed namespaces.

* **Tests**
  * Expanded unit tests covering selector/glob filtering, combined flows, and allowlist gating.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -26,18 +26,19 @@ import (
 
 // Out of box defaults
 const (
-	COLLECTOR_API_VERSION      = "2.15.0"
-	DEFAULT_AGGREGATOR_URL     = "https://localhost:3010" // this will be deprecated in the future
-	DEFAULT_AGGREGATOR_HOST    = "https://localhost"
-	DEFAULT_AGGREGATOR_PORT    = "3010"
-	DEFAULT_CLUSTER_NAME       = "local-cluster"
-	DEFAULT_POD_NAMESPACE      = "open-cluster-management"
-	DEFAULT_HEARTBEAT_MS       = 300000 // 5 min
-	DEFAULT_MAX_BACKOFF_MS     = 600000 // 10 min
-	DEFAULT_REDISCOVER_RATE_MS = 60000  // 1 min
-	DEFAULT_REPORT_RATE_MS     = 5000   // 5 seconds
-	DEFAULT_RETRY_JITTER_MS    = 5000   // 5 seconds
-	DEFAULT_RUNTIME_MODE       = "production"
+	COLLECTOR_API_VERSION          = "2.15.0"
+	DEFAULT_AGGREGATOR_URL         = "https://localhost:3010" // this will be deprecated in the future
+	DEFAULT_AGGREGATOR_HOST        = "https://localhost"
+	DEFAULT_AGGREGATOR_PORT        = "3010"
+	DEFAULT_CLUSTER_NAME           = "local-cluster"
+	DEFAULT_POD_NAMESPACE          = "open-cluster-management"
+	DEFAULT_HEARTBEAT_MS           = 300000 // 5 min
+	DEFAULT_MAX_BACKOFF_MS         = 600000 // 10 min
+	DEFAULT_REDISCOVER_RATE_MS     = 60000  // 1 min
+	DEFAULT_REPORT_RATE_MS         = 5000   // 5 seconds
+	DEFAULT_NS_FILTER_CACHE_TTL_MS = 60000  // 1 min
+	DEFAULT_RETRY_JITTER_MS        = 5000   // 5 seconds
+	DEFAULT_RUNTIME_MODE           = "production"
 )
 
 // Configuration options for the search-collector.
@@ -57,6 +58,7 @@ type Config struct {
 	HTTPTimeout                   int          `env:"HTTP_TIMEOUT"`                    // Timeout for http server connections. Default: 5 min
 	KubeConfig                    string       `env:"KUBECONFIG"`                      // Local kubeconfig path
 	MaxBackoffMS                  int          `env:"MAX_BACKOFF_MS"`                  // Maximum backoff in ms to wait after error
+	NSFilterCacheTTLMS            int          `env:"NS_FILTER_CACHE_TTL_MS"`          // TTL(ms) for the namespace filter cache
 	PodNamespace                  string       `env:"POD_NAMESPACE"`                   // The namespace of this pod
 	RediscoverRateMS              int          `env:"REDISCOVER_RATE_MS"`              // Interval(ms) between CRD discovery syncs
 	RetryJitterMS                 int          `env:"RETRY_JITTER_MS"`                 // Random jitter added to backoff wait.
@@ -103,6 +105,7 @@ func InitConfig() {
 
 	setDefaultInt(&Cfg.HeartbeatMS, "HEARTBEAT_MS", DEFAULT_HEARTBEAT_MS)
 	setDefaultInt(&Cfg.MaxBackoffMS, "MAX_BACKOFF_MS", DEFAULT_MAX_BACKOFF_MS)
+	setDefaultInt(&Cfg.NSFilterCacheTTLMS, "NS_FILTER_CACHE_TTL_MS", DEFAULT_NS_FILTER_CACHE_TTL_MS)
 	setDefaultInt(&Cfg.RediscoverRateMS, "REDISCOVER_RATE_MS", DEFAULT_REDISCOVER_RATE_MS)
 	setDefaultInt(&Cfg.ReportRateMS, "REPORT_RATE_MS", DEFAULT_REPORT_RATE_MS)
 	setDefaultInt(&Cfg.RetryJitterMS, "RETRY_JITTER_MS", DEFAULT_RETRY_JITTER_MS)
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -36,7 +36,7 @@ const (
 	DEFAULT_MAX_BACKOFF_MS         = 600000 // 10 min
 	DEFAULT_REDISCOVER_RATE_MS     = 60000  // 1 min
 	DEFAULT_REPORT_RATE_MS         = 5000   // 5 seconds
-	DEFAULT_NS_FILTER_CACHE_TTL_MS = 60000  // 1 min
+	DEFAULT_NS_FILTER_CACHE_TTL_MS = 300000 // 5 min
 	DEFAULT_RETRY_JITTER_MS        = 5000   // 5 seconds
 	DEFAULT_RUNTIME_MODE           = "production"
 )
--- a/pkg/informer/informer.go
+++ b/pkg/informer/informer.go
@@ -131,7 +131,12 @@ func filterByGlobs(nsList *v1.NamespaceList, nsSelector *v1alpha1.NamespaceSelec
 		if len(nsSelector.Include) > 0 {
 			matched := false
 			for _, pattern := range nsSelector.Include {
-				if ok, _ := filepath.Match(pattern, name); ok {
+				ok, err := filepath.Match(pattern, name)
+				if err != nil {
+					klog.Warningf("Invalid include glob pattern %q: %v", pattern, err)
+					continue
+				}
+				if ok {
 					matched = true
 					break
 				}
@@ -145,7 +150,12 @@ func filterByGlobs(nsList *v1.NamespaceList, nsSelector *v1alpha1.NamespaceSelec
 		if len(nsSelector.Exclude) > 0 {
 			excluded := false
 			for _, pattern := range nsSelector.Exclude {
-				if ok, _ := filepath.Match(pattern, name); ok {
+				ok, err := filepath.Match(pattern, name)
+				if err != nil {
+					klog.Warningf("Invalid exclude glob pattern %q: %v", pattern, err)
+					continue
+				}
+				if ok {
 					excluded = true
 					break
 				}
--- a/pkg/informer/informer.go
+++ b/pkg/informer/informer.go
@@ -370,6 +370,10 @@ func isNamespaceAllowed(allowedNSMap map[string]bool, namespace string) bool {
 		return true
 	}
 
+	if allowedNSMap == nil { // error in processing allowedNamespaceMap: collect everywhere
+		return true
+	}
+
 	_, ok := allowedNSMap[namespace]
 	return ok
 }
--- a/pkg/informer/informer.go
+++ b/pkg/informer/informer.go
@@ -6,6 +6,7 @@ package informer
 import (
 	"context"
 	"path/filepath"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -20,6 +21,36 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 )
+
+type namespaceCache struct {
+	mu        sync.RWMutex
+	allowed   map[string]bool
+	expiresAt time.Time
+	ttl       time.Duration
+}
+
+// get returns the cached namespace map, refreshing it if the TTL has expired.
+func (c *namespaceCache) get() map[string]bool {
+	c.mu.RLock()
+	if time.Now().Before(c.expiresAt) {
+		defer c.mu.RUnlock()
+		return c.allowed
+	}
+	c.mu.RUnlock()
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	// Double-check: another goroutine may have refreshed while we waited for the write lock.
+	if time.Now().Before(c.expiresAt) {
+		return c.allowed
+	}
+
+	c.allowed = resolveCollectNamespaces()
+	c.expiresAt = time.Now().Add(c.ttl)
+	return c.allowed
+}
+
+var nsCache = &namespaceCache{ttl: 2 * time.Minute}
 
 // GenericInformer ...
 type GenericInformer struct {
@@ -71,7 +102,7 @@ func (inform *GenericInformer) Run(ctx context.Context) {
 				inform.client = config.GetDynamicClient()
 			}
 
-			collectNamespaces := getCollectNamespaces()
+			collectNamespaces := nsCache.get()
 
 			err := inform.listAndResync(collectNamespaces)
 			if err == nil {
@@ -171,9 +202,9 @@ func filterByGlobs(nsList *v1.NamespaceList, nsSelector *v1alpha1.NamespaceSelec
 	return result
 }
 
-// getCollectNamespaces resolves the CollectorConfig namespaceSelector to a flat list of namespace names.
-// Follows the config-policy-controller pattern: labels first, then include globs, then exclude globs.
-func getCollectNamespaces() map[string]bool {
+// resolveCollectNamespaces fetches the CollectorConfig and resolves the namespaceSelector
+// to a flat map of allowed namespace names. Called by nsCache.get() when the TTL expires.
+func resolveCollectNamespaces() map[string]bool {
 	if !config.Cfg.FeatureConfigurableCollection {
 		return nil
 	}
--- a/pkg/informer/informer.go
+++ b/pkg/informer/informer.go
@@ -5,17 +5,18 @@ package informer
 
 import (
 	"context"
+	"path/filepath"
 	"sync/atomic"
 	"time"
 
 	"github.com/stolostron/search-collector/pkg/config"
+	"github.com/stolostron/search-v2-operator/api/v1alpha1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/klog/v2"
-
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-
 	"k8s.io/client-go/dynamic"
+	"k8s.io/klog/v2"
 )
 
 // GenericInformer ...
@@ -68,10 +69,12 @@ func (inform *GenericInformer) Run(ctx context.Context) {
 				inform.client = config.GetDynamicClient()
 			}
 
-			err := inform.listAndResync()
+			collectNamespaces := getCollectNamespaces()
+
+			err := inform.listAndResync(collectNamespaces)
 			if err == nil {
 				inform.initialized.Store(true)
-				inform.watch(ctx.Done())
+				inform.watch(ctx.Done(), collectNamespaces)
 			}
 		}
 	}
@@ -89,9 +92,110 @@ func newUnstructured(kind, uid string) *unstructured.Unstructured {
 	}
 }
 
+// getCollectNamespaces resolves the CollectorConfig namespaceSelector to a flat list of namespace names.
+// Follows the config-policy-controller pattern: labels first, then include globs, then exclude globs.
+func getCollectNamespaces() map[string]bool {
+	if !config.Cfg.FeatureConfigurableCollection {
+		return nil
+	}
+
+	unstructuredConfig, err := config.GetDynamicClient().Resource(schema.GroupVersionResource{
+		Group:    "search.open-cluster-management.io",
+		Version:  "v1alpha1",
+		Resource: "collectorconfigs",
+	}).Namespace(config.Cfg.PodNamespace).Get(context.Background(), "collector-config", metav1.GetOptions{})
+	if err != nil {
+		klog.Infof("Could not load collector-config resource: %v. Using default config only.", err)
+		return nil
+	}
+
+	var collectorConfig v1alpha1.CollectorConfig
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(unstructuredConfig.Object, &collectorConfig); err != nil {
+		klog.Warningf("Could not convert collector-config to typed object: %v. Using default config only.", err)
+		return nil
+	}
+
+	// No collectNamespaces or namespaceSelector configured — collect everything
+	if collectorConfig.Spec.CollectNamespaces == nil || collectorConfig.Spec.CollectNamespaces.NamespaceSelector == nil {
+		return nil
+	}
+	nsSelector := collectorConfig.Spec.CollectNamespaces.NamespaceSelector
+
+	// Nothing specified — collect everything
+	if len(nsSelector.Include) == 0 && len(nsSelector.Exclude) == 0 &&
+		len(nsSelector.MatchLabels) == 0 && len(nsSelector.MatchExpressions) == 0 {
+		return nil
+	}
+
+	// Build a label selector from matchLabels and matchExpressions
+	labelSelector := ""
+	if len(nsSelector.MatchLabels) > 0 || len(nsSelector.MatchExpressions) > 0 {
+		ls := &metav1.LabelSelector{
+			MatchLabels:      nsSelector.MatchLabels,
+			MatchExpressions: nsSelector.MatchExpressions,
+		}
+		selector, err := metav1.LabelSelectorAsSelector(ls)
+		if err != nil {
+			klog.Warningf("Error parsing namespace label selector: %v. Skipping label filtering.", err)
+		} else {
+			labelSelector = selector.String()
+		}
+	}
+
+	// List namespaces filtered by labels
+	kubeClient := config.GetKubeClient(config.GetKubeConfig())
+	nsList, err := kubeClient.CoreV1().Namespaces().List(context.Background(), metav1.ListOptions{
+		LabelSelector: labelSelector,
+	})
+	if err != nil {
+		klog.Warningf("Error listing namespaces: %v. Skipping namespace filtering.", err)
+		return nil
+	}
+
+	// Apply include/exclude filepath globs
+	result := make(map[string]bool, 0)
+	for _, ns := range nsList.Items {
+		name := ns.Name
+
+		// Include filter: if include list is specified, namespace must match at least one pattern
+		if len(nsSelector.Include) > 0 {
+			matched := false
+			for _, pattern := range nsSelector.Include {
+				if ok, _ := filepath.Match(pattern, name); ok {
+					matched = true
+					break
+				}
+			}
+			if !matched {
+				continue
+			}
+		}
+
+		// Exclude filter: if namespace matches any exclude pattern, skip it
+		if len(nsSelector.Exclude) > 0 {
+			excluded := false
+			for _, pattern := range nsSelector.Exclude {
+				if ok, _ := filepath.Match(pattern, name); ok {
+					excluded = true
+					break
+				}
+			}
+			if excluded {
+				continue
+			}
+		}
+
+		// Add namespace to map that met labels, expressions, includes, and excludes
+		result[name] = true
+	}
+
+	klog.V(3).Infof("Resolved collectNamespaces to %d namespaces: %v", len(result), result)
+	return result
+}
+
 // List current resources and fires ADDED events. Then sync the current state with the previous
 // state and delete any resources that are still in our cache, but no longer exist in the cluster.
-func (inform *GenericInformer) listAndResync() error {
+func (inform *GenericInformer) listAndResync(collectNamespaces map[string]bool) error {
 
 	// Keep track of new resources added to consolidate against the previous state.
 	newResourceIndex := make(map[string]string)
@@ -107,8 +211,12 @@ func (inform *GenericInformer) listAndResync() error {
 			return listError
 		}
 
-		// Add all resources.
+		// Add all resources filtered by namespace
 		for i := range resources.Items {
+			if !isNamespaceAllowed(collectNamespaces, resources.Items[i].GetNamespace()) {
+				continue
+			}
+
 			klog.V(5).Infof("KIND: %s UUID: %s, ResourceVersion: %s",
 				inform.gvr.Resource, resources.Items[i].GetUID(), resources.Items[i].GetResourceVersion())
 			inform.AddFunc(&resources.Items[i])
@@ -140,7 +248,7 @@ func (inform *GenericInformer) listAndResync() error {
 }
 
 // Watch resources and process events.
-func (inform *GenericInformer) watch(stopper <-chan struct{}) {
+func (inform *GenericInformer) watch(stopper <-chan struct{}, collectNamespaces map[string]bool) {
 	watch, watchError := inform.client.Resource(inform.gvr).Watch(context.TODO(), metav1.ListOptions{})
 	if watchError != nil {
 		klog.Warningf("Error watching resources for %s.  Error: %s", inform.gvr.String(), watchError)
@@ -171,6 +279,11 @@ func (inform *GenericInformer) watch(stopper <-chan struct{}) {
 						inform.gvr.Resource)
 					continue
 				}
+
+				if !isNamespaceAllowed(collectNamespaces, obj.GetNamespace()) {
+					continue
+				}
+
 				inform.AddFunc(obj)
 				inform.resourceIndex[string(obj.GetUID())] = obj.GetResourceVersion()
 
@@ -183,6 +296,10 @@ func (inform *GenericInformer) watch(stopper <-chan struct{}) {
 					continue
 				}
 
+				if !isNamespaceAllowed(collectNamespaces, obj.GetNamespace()) {
+					continue
+				}
+
 				inform.UpdateFunc(nil, obj)
 				inform.resourceIndex[string(obj.GetUID())] = obj.GetResourceVersion()
 
@@ -192,6 +309,10 @@ func (inform *GenericInformer) watch(stopper <-chan struct{}) {
 				if !ok {
 					klog.Warningf("Error converting %s event.Object to unstructured.Unstructured on DELETED event",
 						inform.gvr.Resource)
+					continue
+				}
+
+				if !isNamespaceAllowed(collectNamespaces, obj.GetNamespace()) {
 					continue
 				}
 
@@ -221,4 +342,13 @@ func (inform *GenericInformer) WaitUntilInitialized(timeout time.Duration) {
 		}
 		time.Sleep(time.Duration(10) * time.Millisecond)
 	}
+}
+
+func isNamespaceAllowed(allowedNSMap map[string]bool, namespace string) bool {
+	if !config.Cfg.FeatureConfigurableCollection {
+		return true
+	}
+
+	_, ok := allowedNSMap[namespace]
+	return ok
 }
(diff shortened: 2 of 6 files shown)


pkg/informer/informer.go
```diff
@@ -222,3 +360,12 @@ func (inform *GenericInformer) WaitUntilInitialized(timeout time.Duration) {
...
 }
+
+func isNamespaceAllowed(allowedNSMap map[string]bool, namespace string) bool {
+	if !config.Cfg.FeatureConfigurableCollection {
+		return true
+	}
+
+	_, ok := allowedNSMap[namespace]
+	return ok
+}
```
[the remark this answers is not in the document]
smcavey: So if there are no namespaces permitted to collect resources for, you would suggest to collect from every resource? Process that suggestion again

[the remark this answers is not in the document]
smcavey: But if there's no filters configured:
```
	nsList, err := kubeClient.CoreV1().Namespaces().List(context.Background(), metav1.ListOptions{
		LabelSelector: labelSelector,
	})
```
in filterBySelectors should return every namespace, and then in filterByGlobs, len(nsSelector.Include) == 0 and len(nsSelector.Exclude) == 0, resulting in every namespace getting set in the map. See unit test TestFilterByEmptySelectors() and tell me I'm still wrong.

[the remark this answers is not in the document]
smcavey: Gotcha, a miscommunication

pkg/informer/informer.go
```diff
@@ -89,9 +94,125 @@ func newUnstructured(kind, uid string) *unstructured.Unstructured {
...
+				continue
+			}
+		}
+
+		// Exclude filter: if namespace matches any exclude pattern, skip it
+		if len(nsSelector.Exclude) > 0 {
+			excluded := false
+			for _, pattern := range nsSelector.Exclude {
+				if ok, _ := filepath.Match(pattern, name); ok {
+					excluded = true
```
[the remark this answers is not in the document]
smcavey: I could perhaps see logging that error. But the source code for filepath.Match https://cs.opensource.google/go/go/+/refs/tags/go1.26.2:src/path/filepath/match.go;l=46-89 returns false, err when err != nil. filterBySelectors presents this function with a full list of namespaces. If there's errors in the filepath.Match calls on Include, we don't add it to the list, if there's errors in filepath.Match calls on Exclude, we don't remove it from the list. I think the better solution is the klog.Warning on error and add validation in the CRD validating webhook to ensure the path is good

pkg/informer/informer.go
```diff
@@ -89,9 +94,125 @@ func newUnstructured(kind, uid string) *unstructured.Unstructured {
...
+		ls := &metav1.LabelSelector{
+			MatchLabels:      nsSelector.MatchLabels,
+			MatchExpressions: nsSelector.MatchExpressions,
+		}
+		selector, err := metav1.LabelSelectorAsSelector(ls)
+		if err != nil {
+			klog.Warningf("Error parsing namespace label selector: %v. Skipping label filtering.", err)
+		} else {
+			labelSelector = selector.String()
+		}
```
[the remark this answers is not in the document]
smcavey: Like we discussed https://github.com/stolostron/search-collector/pull/866#discussion_r3188393817 we are logging, I think we move this validation to the webhook as well as opposed to failing. All of this stuff is feature flagged, the default behavior of the system would be to collect all namespaces. If there's an error here I'd like to keep that same functionality, which is what's implemented. But we could do further validation on the webhook.

pkg/informer/informer.go
```diff
@@ -5,19 +5,53 @@ package informer
...
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog/v2"
 )
 
+type namespaceCache struct {
```
jlpadilla: The struct name can be confused with the existing namespaces cache. 
Does `namespaceFilterCache` makes more sense here?

pkg/informer/informer.go
```diff
@@ -5,19 +5,53 @@ package informer
...
+	if time.Now().Before(c.expiresAt) {
+		return c.allowed
+	}
+
+	c.allowed = resolveCollectNamespaces()
+	c.expiresAt = time.Now().Add(c.ttl)
+	return c.allowed
+}
+
+var nsCache = &namespaceCache{ttl: 2 * time.Minute}
```
jlpadilla: Make this ttl configurable.

pkg/informer/informer.go
```diff
@@ -89,9 +125,135 @@ func newUnstructured(kind, uid string) *unstructured.Unstructured {
 	}
 }
```
jlpadilla, line 89: Can you move these new functions to a new file?  Maybe `namespaceFilter.go`?

pkg/informer/informer.go
jlpadilla: or maybe `collectNamespacesCache`.

pkg/informer/informer.go
```diff
@@ -89,9 +125,135 @@ func newUnstructured(kind, uid string) *unstructured.Unstructured {
...
+	}
+	return result
+}
+
+// resolveCollectNamespaces fetches the CollectorConfig and resolves the namespaceSelector
+// to a flat map of allowed namespace names. Called by nsCache.get() when the TTL expires.
+func resolveCollectNamespaces() map[string]bool {
+	if !config.Cfg.FeatureConfigurableCollection {
+		return nil
+	}
```
jlpadilla: I'd move this check earlier, in the `get()` function or before.

pkg/informer/informer.go
```diff
@@ -89,9 +125,135 @@ func newUnstructured(kind, uid string) *unstructured.Unstructured {
...
+
+	unstructuredConfig, err := config.GetDynamicClient().Resource(schema.GroupVersionResource{
+		Group:    "search.open-cluster-management.io",
+		Version:  "v1alpha1",
+		Resource: "collectorconfigs",
+	}).Namespace(config.Cfg.PodNamespace).Get(context.Background(), "collector-config", metav1.GetOptions{})
+	if err != nil {
+		klog.Infof("Could not load collector-config resource: %v. Using default config only.", err)
```
jlpadilla: Should this log say "Will collect data from all namespaces" ?

jlpadilla: This should be a klog.Warning

pkg/informer/informer.go
```diff
@@ -89,9 +125,135 @@ func newUnstructured(kind, uid string) *unstructured.Unstructured {
...
+		return nil
+	}
+
+	var collectorConfig v1alpha1.CollectorConfig
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(unstructuredConfig.Object, &collectorConfig); err != nil {
+		klog.Warningf("Could not convert collector-config to typed object: %v. Using default config only.", err)
```
jlpadilla: Log should say  "Will collect data from all namespaces"

pkg/informer/informer.go
```diff
@@ -89,9 +125,135 @@ func newUnstructured(kind, uid string) *unstructured.Unstructured {
...
+		return nil
+	}
+
+	// No collectNamespaces or namespaceSelector configured — collect everywhere
+	if collectorConfig.Spec.CollectNamespaces == nil || collectorConfig.Spec.CollectNamespaces.NamespaceSelector == nil {
+		return nil
```
jlpadilla: Log this decision. Maybe `klog.V(2).Info(...)`

pkg/informer/informer.go
```diff
@@ -89,9 +125,135 @@ func newUnstructured(kind, uid string) *unstructured.Unstructured {
...
+	}
+	nsSelector := collectorConfig.Spec.CollectNamespaces.NamespaceSelector
+
+	// Nothing specified — collect everywhere
+	if len(nsSelector.Include) == 0 && len(nsSelector.Exclude) == 0 &&
+		len(nsSelector.MatchLabels) == 0 && len(nsSelector.MatchExpressions) == 0 {
+		return nil
```
jlpadilla: Log decision.

pkg/informer/informer.go
```diff
@@ -89,9 +125,135 @@ func newUnstructured(kind, uid string) *unstructured.Unstructured {
...
+	}
+
+	// List namespaces filtered by labelSelectors and matchExpressions
+	nsList, err := filterBySelectors(config.GetKubeClient(config.GetKubeConfig()), nsSelector)
+	if err != nil {
+		klog.Warningf("Error listing namespaces: %v. Skipping namespace filtering.", err)
```
jlpadilla:
```suggestion
		klog.Warningf("Error listing namespaces by labelSelector or matchExpressions. Skipping namespace filtering. %v.", err)
```

pkg/informer/informer.go
```diff
@@ -171,6 +337,11 @@ func (inform *GenericInformer) watch(stopper <-chan struct{}) {
 						inform.gvr.Resource)
 					continue
 				}
+
+				if !isNamespaceAllowed(collectNamespaces, obj.GetNamespace()) {
```
jlpadilla: 1. I don't think the data in `collectNamespaces` gets updated after the watch is started.
2. If it does, then we need to add a lock while we are reading from it.

pkg/informer/informer.go
```diff
@@ -171,9 +173,19 @@ func (inform *GenericInformer) watch(stopper <-chan struct{}) {
...
+
+				if !isNamespaceAllowed(obj.GetNamespace()) {
+					continue
+				}
+
 				inform.AddFunc(obj)
 				inform.resourceIndex[string(obj.GetUID())] = obj.GetResourceVersion()
 
+				// Namespace changes affect which resources pass the namespace filter.
+				if inform.gvr.Resource == "namespaces" {
```
jlpadilla: Also check apigroup, just in case anyone defines a CRD with the same name.

pkg/informer/informer.go
```diff
@@ -183,9 +195,18 @@ func (inform *GenericInformer) watch(stopper <-chan struct{}) {
...
 				}
 
+				if !isNamespaceAllowed(obj.GetNamespace()) {
+					continue
+				}
+
 				inform.UpdateFunc(nil, obj)
 				inform.resourceIndex[string(obj.GetUID())] = obj.GetResourceVersion()
 
+				// Label changes on namespaces can affect label selector results.
```
jlpadilla: This scenario is more complex.
- We'll need to remove data from any namespace that no longer matches the selector.
- We'll need to add data for any namespace that was not being collected before.

This is likely complex enough for a separate PR.

pkg/informer/informer.go
```diff
@@ -195,9 +216,18 @@ func (inform *GenericInformer) watch(stopper <-chan struct{}) {
...
+				if !isNamespaceAllowed(obj.GetNamespace()) {
+					continue
+				}
+
 				inform.DeleteFunc(obj)
 				delete(inform.resourceIndex, string(obj.GetUID()))
 
+				// Namespace deletion affects the allowed namespace set.
+				if inform.gvr.Resource == "namespaces" {
+					nsFilterCache.invalidate()
```
jlpadilla: We could simply remove the namespace from the cache, no need to rebuild.

pkg/informer/informer.go
```diff
@@ -195,9 +216,18 @@ func (inform *GenericInformer) watch(stopper <-chan struct{}) {
 					continue
 				}
 
+				if !isNamespaceAllowed(obj.GetNamespace()) {
+					continue
+				}
```
jlpadilla: There's a very small chance there could be a problem if we receive events out of order.
**Scenario:**
    1. Namespace is deleted.
    2. We remove namespace from cache.
    3. A DELETE event for a resource in the namespace arrives late.

**Alternative:** We could always process delete events. Just need to double check it doesn't cause a problem in the delete logic if the resource is not in our data.

pkg/informer/informer.go
```diff
@@ -183,9 +195,18 @@ func (inform *GenericInformer) watch(stopper <-chan struct{}) {
...
+
 				inform.UpdateFunc(nil, obj)
 				inform.resourceIndex[string(obj.GetUID())] = obj.GetResourceVersion()
 
+				// Label changes on namespaces can affect label selector results.
+				if inform.gvr.Resource == "namespaces" {
```
jlpadilla: Here we also need to watch and invalidate if the CollectorConfig resource changes.

pkg/informer/informer.go
smcavey: Separate PR 👍

pkg/informer/informer.go
```diff
@@ -171,9 +173,19 @@ func (inform *GenericInformer) watch(stopper <-chan struct{}) {
...
+					continue
+				}
+
 				inform.AddFunc(obj)
 				inform.resourceIndex[string(obj.GetUID())] = obj.GetResourceVersion()
 
+				// Namespace changes affect which resources pass the namespace filter.
+				if inform.gvr.Resource == "namespaces" {
+					nsFilterCache.invalidate()
+				}
```
jlpadilla: When the namespace informer is first started, we receive events for all existing namespaces. This will trigger too many unnecessary invalidations.

**Options:**
- We could check if the namespace is already in the filter, no need to invalidate for ADD. There's still a problem for all namespaces that didn't match the filter.

pkg/informer/informer.go
smcavey: Would like to handle this with https://redhat.atlassian.net/browse/ACM-20047. I apparently naively thought it would be simple to just invalidate and reprocess the namespace filter cache on each case. If namespace it deleted, we shouldn't be getting anymore events for that namespace, anyway. I'm inclined to leave the invalidation call on the case ADDED alone for simplicity. MODIFIED is too complex, DELETED is unnecessary

pkg/informer/namespaceFilter.go
```diff
@@ -0,0 +1,206 @@
...
+	c.expiresAt = time.Now().Add(time.Duration(config.Cfg.NSFilterCacheTTLMS) * time.Millisecond)
+	return c.allowed
+}
+
+// invalidate forces the next get() call to refresh the cache.
+func (c *namespaceFilterCache) invalidate() {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.expiresAt = time.Time{} // zero time is always before now
+}
```
jlpadilla: I'm thinking we should trigger the cache refresh instead of waiting for the next event.

pkg/informer/namespaceFilter.go
```diff
@@ -0,0 +1,206 @@
...
+	if namespace == "" { // cluster-scoped resources don't have namespace
+		return true
+	}
+
+	allowedNSMap := nsFilterCache.get()
+	if allowedNSMap == nil { // error in processing allowedNamespaceMap: collect everywhere
+		return true
+	}
+
+	_, ok := allowedNSMap[namespace]
```
jlpadilla, line 78: I think we need a lock to read here because this is a reference to the original map.

jlpadilla: In that case it's better to add
```go
func (c *namespaceFilterCache) isAllowed(ns string) bool
```

smcavey: I think in this case when the cache is invalidated there's a small chance that after we call get() and then begin to read from the map here there's a chance that we end up reading from a different map if another goroutine recomputes the map, I don't think it's strictly necessary but that is probably better practice

pkg/informer/namespaceFilter.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+// filterByGlobs applies include/exclude filepath glob patterns to narrow a namespace list.
+func filterByGlobs(nsList *v1.NamespaceList, nsSelector *v1alpha1.NamespaceSelector) map[string]bool {
+	result := make(map[string]bool, 0)
+	for _, ns := range nsList.Items {
+		name := ns.Name
+
+		// Include filter: if include list is specified, namespace must match at least one pattern
+		if len(nsSelector.Include) > 0 {
```
jlpadilla, line 120: The order we process include/exclude is important. I wonder if there's a precedence with policy.

For example:
```
exclude: openshift*
include: openshift-console
```
With the current implementation `oenshift-console` won't be included.

smcavey: https://github.com/stolostron/config-policy-controller/blob/main/pkg/common/pattern_util.go#L21-L64

pkg/informer/namespaceFilter.go
```diff
@@ -0,0 +1,206 @@
...
+
+	// List namespaces filtered by labelSelectors and matchExpressions
+	nsList, err := filterBySelectors(config.GetKubeClient(config.GetKubeConfig()), nsSelector)
+	if err != nil {
+		klog.Warningf("Error listing namespaces by matchLabel or matchExpression: %v. Skipping namespace filtering.", err)
+		return nil
+	}
+
+	// Filter namespaces by include and exclude namespace globs
+	result := filterByGlobs(nsList, nsSelector)
```
jlpadilla, line 213: Need more clarity about how we expect the combination of `selector + include/exclude` to work together.

With the current implementation we first apply the selector and then include/exclude.  This could be confusing if not defined clearly.

Example:
```
MatchLabels:
  - name=foo
Include:
  - bar
```
I'd probably expect `foo and bar` to get collected, but I don't think we are collecting either.

pkg/informer/namespaceFilter.go
jlpadilla: It can be worse because there's a chance that read and write happen exactly at the same moment and 💥

pkg/informer/namespaceFilter.go
jlpadilla: I'm good matching Policy for consistency. Let's document it well because I was expecting something slightly different.

pkg/informer/namespaceFilter.go
smcavey: Renamed to `regenerate()` to immediately rebuild the cache and reset TTL

pkg/informer/namespaceFilter.go
smcavey: Updated func comment:
```
// resolveCollectNamespaces fetches the CollectorConfig and resolves the namespaceSelector
// to a flat map of allowed namespace names. Called by nsFilterCache.get() when the cache has expired.
// three-step process:
// 1. build a selector that ANDs matchLabels and matchExpressions, returning set of all namespaces that match
// 2. check if include list namespaces are in the list, and if so,
// 3. check if they should then be removed via the excludes list
func resolveCollectNamespaces() map[string]bool {
```

pkg/informer/informer.go
smcavey: Added comment:
```
				// FUTURE: Better handing on namespace additions
				//   on startup and when we restart informers we will get a lot of namespaces being ADDED
				//   results in lots of invalidation and computation of namespace filter cache
				//   for example check if namespace is already in filter and skip invalidation on ADDED
				//   https://github.com/stolostron/search-collector/pull/866#discussion_r3196919607
				if !nsFilterCache.isNamespaceAllowed(obj.GetNamespace()) {
```

pkg/informer/informer.go
smcavey: Added comment:
```
				// FUTURE: Better handling on namespace modifications
				//   for example invalidate the namespace filter cache on CollectorConfig changes
				//   https://github.com/stolostron/search-collector/pull/866#discussion_r3196850432
				if !nsFilterCache.isNamespaceAllowed(obj.GetNamespace()) {
```

pkg/informer/informer.go
smcavey: Added comment:
```
				// FUTURE: better handling on namespace deletion
				//   for example remove the namespace key from the map
				//   https://github.com/stolostron/search-collector/pull/866#discussion_r3196773820
```

pkg/informer/informer.go
smcavey: Checking group:
```
				if inform.gvr.Resource == "namespaces" && inform.gvr.Group == "" {
					nsFilterCache.regenerate()
				}
```

pkg/informer/informer.go
```diff
@@ -195,6 +224,17 @@ func (inform *GenericInformer) watch(stopper <-chan struct{}) {
...
+				// FUTURE: better handling on namespace deletion
+				//   for example remove the namespace key from the map
+				//   https://github.com/stolostron/search-collector/pull/866#discussion_r3196773820
+				if !nsFilterCache.isNamespaceAllowed(obj.GetNamespace()) {
+					continue
+				}
+				// Namespace deletions affect which resources pass the namespace filter.
+				if inform.gvr.Resource == "namespaces" && inform.gvr.Group == "" {
+					nsFilterCache.regenerate()
+				}
```
jlpadilla: Don't regenerate on deletion.

pkg/config/config.go
```diff
@@ -26,18 +26,19 @@ import (
...
+	DEFAULT_AGGREGATOR_URL         = "https://localhost:3010" // this will be deprecated in the future
+	DEFAULT_AGGREGATOR_HOST        = "https://localhost"
+	DEFAULT_AGGREGATOR_PORT        = "3010"
+	DEFAULT_CLUSTER_NAME           = "local-cluster"
+	DEFAULT_POD_NAMESPACE          = "open-cluster-management"
+	DEFAULT_HEARTBEAT_MS           = 300000 // 5 min
+	DEFAULT_MAX_BACKOFF_MS         = 600000 // 10 min
+	DEFAULT_REDISCOVER_RATE_MS     = 60000  // 1 min
+	DEFAULT_REPORT_RATE_MS         = 5000   // 5 seconds
+	DEFAULT_NS_FILTER_CACHE_TTL_MS = 60000  // 1 min
```
jlpadilla: Increase to 5 minutes.

This cache is updated from events, so this TTL is just a fallback.

pkg/informer/namespaceFilter.go
```diff
@@ -0,0 +1,213 @@
...
+	if err != nil {
+		klog.Warningf("Error listing namespaces: %v. Skipping namespace filtering.", err)
+		return nil, err
+	}
+
+	return nsList, nil
+}
+
+// filterByGlobs applies include/exclude filepath glob patterns to narrow a namespace list.
+func filterByGlobs(nsList *v1.NamespaceList, nsSelector *v1alpha1.NamespaceSelector) map[string]bool {
```
jlpadilla, line 114:
```suggestion
// filterByGlobs applies include/exclude filepath glob patterns to narrow a namespace list.
// This should remain consistent with the GRC implementation.
// 
https://github.com/stolostron/config-policy-controller/blob/main/pkg/common/pattern_util.go#L21-L64
func filterByGlobs(nsList *v1.NamespaceList, nsSelector *v1alpha1.NamespaceSelector) map[string]bool {
```

pkg/informer/namespaceFilter.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	return result
+}
+
+// resolveCollectNamespaces fetches the CollectorConfig and resolves the namespaceSelector
+// to a flat map of allowed namespace names. Called by nsFilterCache.get() when the cache has expired.
+// three-step process:
+// 1. build a selector that ANDs matchLabels and matchExpressions, returning set of all namespaces that match
+// 2. check if include list namespaces are in the list, and if so,
+// 3. check if they should then be removed via the excludes list
```
jlpadilla:
```suggestion
// resolveCollectNamespaces fetches the CollectorConfig and resolves the namespaceSelector
// to a flat map of allowed namespace names. Called by nsFilterCache.get() when the cache has expired.
// This logic matches the GRC namespace selector. 
// Three-step process:
//   1. build a selector that ANDs matchLabels and matchExpressions, returning set of all namespaces that match
//   2. check if include list namespaces are in the list, and if so,
//   3. check if they should then be removed via the excludes list
```

pkg/informer/namespaceFilter.go
```diff
@@ -0,0 +1,213 @@
...
+		return nil
+	}
+
+	unstructuredConfig, err := config.GetDynamicClient().Resource(schema.GroupVersionResource{
+		Group:    "search.open-cluster-management.io",
+		Version:  "v1alpha1",
+		Resource: "collectorconfigs",
+	}).Namespace(config.Cfg.PodNamespace).Get(context.Background(), "collector-config", metav1.GetOptions{})
+	if err != nil {
+		klog.Infof("Could not load collector-config resource: %v. Will collect data from all namespaces.", err)
```
jlpadilla:
```suggestion
		klog.Infof("Could not load CollectorConfig resource. Will collect data from all namespaces. Error: %v", err)
```

pkg/informer/namespaceFilter.go
```diff
@@ -0,0 +1,213 @@
...
+		return nil
+	}
+
+	var collectorConfig v1alpha1.CollectorConfig
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(unstructuredConfig.Object, &collectorConfig); err != nil {
+		klog.Warningf("Could not convert collector-config to typed object: %v. Will collect data from all namespaces.", err)
```
jlpadilla:
```suggestion
		klog.Warningf("Could not convert CollectorConfig to typed object. Will collect data from all namespaces. Error: %v", err)
```

pkg/informer/namespaceFilter.go
```diff
@@ -0,0 +1,213 @@
...
+		return nil
+	}
+
+	// No collectNamespaces or namespaceSelector configured — collect everywhere
+	if collectorConfig.Spec.CollectNamespaces == nil || collectorConfig.Spec.CollectNamespaces.NamespaceSelector == nil {
+		klog.V(2).Info("No collectNamespaces or namespaceSelector fields on collector-config. Will collect data from all namespaces.")
```
jlpadilla:
```suggestion
		klog.V(2).Info("No collectNamespaces or namespaceSelector fields on CollectorConfig. Will collect data from all namespaces.")
```

pkg/informer/namespaceFilter.go
```diff
@@ -0,0 +1,213 @@
...
+	if len(nsSelector.Include) == 0 && len(nsSelector.Exclude) == 0 &&
+		len(nsSelector.MatchLabels) == 0 && len(nsSelector.MatchExpressions) == 0 {
+		klog.V(2).Info("No include, exclude, matchLabel, or matchExpression on collector-config. Will collect data from all namespaces.")
+		return nil
+	}
+
+	// List namespaces filtered by labelSelectors and matchExpressions
+	nsList, err := filterBySelectors(config.GetKubeClient(config.GetKubeConfig()), nsSelector)
+	if err != nil {
+		klog.Warningf("Error listing namespaces by matchLabel or matchExpression: %v. Skipping namespace filtering.", err)
```
jlpadilla:
```suggestion
		klog.Warningf("Error listing namespaces by matchLabel or matchExpression. Skipping namespace filtering. Error: %v. ", err)
```

pkg/informer/informer.go
```diff
@@ -107,8 +105,14 @@ func (inform *GenericInformer) listAndResync() error {
...
 		}
 
-		// Add all resources.
+		// Add all resources filtered by namespace
 		for i := range resources.Items {
+			if config.Cfg.FeatureConfigurableCollection {
+				if !nsFilterCache.isNamespaceAllowed(resources.Items[i].GetNamespace()) {
+					continue
+				}
+			}
```
jlpadilla: The feature flag is checked in `isNamespaceAllowed()`

```suggestion
			if !nsFilterCache.isNamespaceAllowed(resources.Items[i].GetNamespace()) {
				continue
			}
```

jlpadilla:
🤖 **I'm an AI agent** providing this code review

## PR Review: ACM-21531 - CollectorConfig NamespaceSelector

**Jira**: [ACM-21531](https://issues.redhat.com/browse/ACM-21531) — Namespace filtering for resource collection

### Overview
This PR implements namespace filtering for the search-collector based on CollectorConfig's NamespaceSelector. The implementation uses a two-stage filtering approach (label selectors → glob patterns) with TTL-based caching to minimize overhead. The feature falls back to "allow all" when disabled or on configuration errors, which maintains backward compatibility.

### ✅ Strengths

1. **Excellent test coverage** (373 lines of tests) — Covers edge cases like empty selectors, nil maps, cluster-scoped resources, wildcard globs, and combined selector+glob scenarios
2. **Proper concurrency handling** — Uses `sync.RWMutex` with double-checked locking pattern in `cache.get()`
3. **Smart fallback behavior** — Returns nil map on errors, which translates to "allow all", preventing data loss on misconfigurations
4. **Clear separation of concerns** — Three distinct functions: `filterBySelectors`, `filterByGlobs`, and `resolveCollectNamespaces`
5. **Good documentation** — Inline comments explain the "why" (e.g., "cluster-scoped resources don't have namespace")
6. **GRC alignment** — Code references and follows GRC's namespace selector implementation

### 🔍 Issues & Suggestions

#### Critical (Must Fix Before Merge)

**1. Race condition in `isNamespaceAllowed` (namespaceFilter.go:60-77)**

```go
// ❌ Current code acquires locks twice
func (c *namespaceFilterCache) isNamespaceAllowed(namespace string) bool {
    // ...
    allowedNSMap := c.get()  // get() acquires locks
    c.mu.RLock()              // Then we acquire read lock again!
    defer c.mu.RUnlock()
    // ...
}
```

The read lock after `c.get()` is redundant and could cause a deadlock if another goroutine is waiting for the write lock. The `allowedNSMap` is a local copy, so no lock is needed after `get()` returns.

**Fix:**
```go
func (c *namespaceFilterCache) isNamespaceAllowed(namespace string) bool {
    if !config.Cfg.FeatureConfigurableCollection {
        return true
    }

    if namespace == "" { // cluster-scoped resources
        return true
    }

    allowedNSMap := c.get()
    // No lock needed here - allowedNSMap is a local reference
    if allowedNSMap == nil { // error in processing: collect everywhere
        return true
    }

    _, ok := allowedNSMap[namespace]
    return ok
}
```

**2. Missing error logging context in `resolveCollectNamespaces` (namespaceFilter.go:167-186)**

When the CollectorConfig cannot be fetched or converted, the error is logged but doesn't include whether this is the first failure or a recurring issue. This makes debugging difficult.

**Suggestion:** Add a counter or log additional context:
```go
klog.Warningf("Could not load CollectorConfig resource (attempt will retry after TTL). Will collect from all namespaces. Error: %v", err)
```

#### Important (Should Fix)

**3. Cache regeneration on every namespace event is expensive (informer.go:178-189, 207-213)**

```go
// Current: regenerates cache on EVERY namespace ADDED/MODIFIED event
if inform.gvr.Resource == "namespaces" && inform.gvr.Group == "" {
    nsFilterCache.regenerate()
}
```

For clusters with frequent namespace churn or during startup (when all existing namespaces fire ADDED events), this will trigger many expensive regenerations. Each regeneration:
- Fetches CollectorConfig from K8s API
- Lists ALL namespaces
- Applies selector/glob filtering

The PR comments acknowledge this (linked future improvements), but this could cause significant performance issues in large clusters.

**Recommendation:** 
- **Option 1**: Add a debounce mechanism (don't regenerate if cache was refreshed within last N seconds)
- **Option 2**: Only regenerate if the namespace change actually affects the filter (check if namespace matches selectors before regenerating)
- **Option 3**: Watch CollectorConfig instead of regenerating on namespace events

For this PR, I'd suggest at least adding:
```go
const minRegenerationInterval = 10 * time.Second

func (c *namespaceFilterCache) regenerate() {
    c.mu.Lock()
    defer c.mu.Unlock()
    
    // Don't regenerate if we just did within the last 10s
    if time.Since(c.expiresAt.Add(-time.Duration(config.Cfg.NSFilterCacheTTLMS) * time.Millisecond)) < minRegenerationInterval {
        klog.V(4).Info("Skipping namespace filter regeneration - recently updated")
        return
    }
    
    c.allowed = resolveCollectNamespaces()
    c.expiresAt = time.Now().Add(time.Duration(config.Cfg.NSFilterCacheTTLMS) * time.Millisecond)
}
```

**4. No observability for filter effectiveness (namespaceFilter.go)**

There are no metrics or logs showing:
- How many namespaces are being filtered
- How many resources are being excluded due to namespace filtering
- Cache hit rate or regeneration frequency

**Suggestion:** Add metrics:
```go
// In resolveCollectNamespaces(), after computing result:
klog.Infof("Namespace filter resolved: %d namespaces allowed out of %d total", len(result), len(nsList.Items))

// In isNamespaceAllowed(), add V(6) logging:
allowed := _, ok := allowedNSMap[namespace]
if !allowed {
    klog.V(6).Infof("Resource filtered: namespace %q not in allowed list", namespace)
}
return allowed
```

**5. Inconsistent namespace event handling (informer.go:224-232)**

The DELETED event handler checks `isNamespaceAllowed` but doesn't regenerate the cache like ADDED/MODIFIED do:

```go
case watch.Deleted:
    // ...
    if !nsFilterCache.isNamespaceAllowed(obj.GetNamespace()) {
        continue
    }
    // Missing: namespace deletion might affect what's allowed
```

If a namespace is deleted, resources from other namespaces might now match the selector (e.g., if the selector excludes a specific namespace and it gets deleted). This is noted in the PR comments as "FUTURE" work, but creates an inconsistency.

**Recommendation:** Either regenerate on DELETE too, or document why it's not needed.

#### Nice to Have (Optional)

**6. Test coverage for cache TTL and concurrency (namespaceFilter_test.go)**

While the tests are comprehensive for filtering logic, they don't test:
- Cache TTL expiration and refresh
- Concurrent access to the cache by multiple goroutines
- Behavior when cache refresh fails midway

**Suggestion:** Add tests like:
```go
func TestCacheTTLExpiration(t *testing.T) {
    // Test that cache actually expires and regenerates after TTL
}

func TestCacheConcurrentAccess(t *testing.T) {
    // Launch multiple goroutines calling isNamespaceAllowed concurrently
}
```

**7. Magic numbers in config (config.go:39)**

```go
DEFAULT_NS_FILTER_CACHE_TTL_MS = 300000 // 5 min
```

5 minutes seems arbitrary. Consider if this should be configurable per-cluster or environment (dev vs prod), or if there's data to support this specific duration.

**8. Potential optimization: skip filter check when nil map (namespaceFilter.go:69-76)**

```go
allowedNSMap := c.get()
if allowedNSMap == nil {
    return true
}
_, ok := allowedNSMap[namespace]
return ok
```

When the map is nil (meaning no filter configured), we still allocate and return it. Minor optimization:

```go
func (c *namespaceFilterCache) get() map[string]bool {
    // ... existing logic ...
    
    // After resolveCollectNamespaces():
    if c.allowed == nil {
        c.expiresAt = time.Now().Add(time.Duration(config.Cfg.NSFilterCacheTTLMS) * time.Millisecond)
        return nil  // Don't bother caching nil - just return it
    }
    
    c.expiresAt = time.Now().Add(time.Duration(config.Cfg.NSFilterCacheTTLMS) * time.Millisecond)
    return c.allowed
}
```

Actually, looking at the code more carefully, this is already handled correctly. Disregard this suggestion.

**9. Consider adding resource version check (informer.go:107-114)**

When filtering resources during `listAndResync`, if the namespace filter changes between when we list resources and when we filter them, we might have stale data. This is unlikely but could cause transient issues.

Not critical for this PR, but worth noting for future improvements.

### 🔒 Security Review

✅ **No security concerns identified:**
- No user input is directly used in glob patterns (patterns come from CollectorConfig CRD)
- Glob pattern errors are handled gracefully (logged and skipped)
- No injection vulnerabilities
- Falls back to permissive behavior on errors (prevents data loss but doesn't leak data)

### 📊 Code Quality

| Aspect | Rating | Notes |
|--------|--------|-------|
| **Correctness** | 🟡 Good | Logic is sound, but race condition in #1 must be fixed |
| **Test Coverage** | 🟢 Excellent | Comprehensive unit tests covering edge cases |
| **Readability** | 🟢 Excellent | Clear structure, good naming, helpful comments |
| **Performance** | 🟡 Good | Cache helps, but regeneration frequency is a concern (#3) |
| **Security** | 🟢 Excellent | No vulnerabilities, proper error handling |
| **Maintainability** | 🟢 Excellent | Well-organized, follows Go idioms |

### 📝 Recommendations

**Before Merge (Critical):**
1. ✅ Fix the redundant lock in `isNamespaceAllowed()` (#1 above)
2. ⚠️ Consider addressing cache regeneration frequency (#3) or document the expected performance impact

**Nice to Have:**
3. Add observability (metrics/logs) for filter effectiveness (#4)
4. Add tests for cache TTL and concurrency (#6)
5. Make DELETE event handling consistent with ADDED/MODIFIED (#5)

### Summary

This is a **well-implemented feature** with excellent test coverage and proper error handling. The two-stage filtering approach (selectors → globs) is clean and matches the GRC implementation, which aids maintainability.

**Critical Issue:** The redundant lock acquisition in `isNamespaceAllowed()` must be fixed before merge.

**Performance Concern:** The cache regeneration on every namespace event could be expensive in large clusters with namespace churn. While the PR comments acknowledge this as "FUTURE" work, consider whether this should be addressed now or if the current implementation is acceptable for the expected usage patterns.

**Recommendation:** ✅ **APPROVE with required fix for #1**. Once the lock issue is fixed, this PR is ready to merge. The performance optimization (#3) can be addressed in a follow-up if monitoring shows it's needed.

---
**Questions for the author:**
1. What is the expected namespace churn rate in target clusters? This informs whether #3 is critical or can be deferred.
2. Was 5 minutes chosen for cache TTL based on specific requirements, or is it arbitrary?
3. Should we add metrics to track filter effectiveness for observability?


[ACM-21531]: https://redhat.atlassian.net/browse/ACM-21531?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

jlpadilla (approved):
/lgtm

pkg/informer/namespaceFilter.go
```diff
@@ -0,0 +1,219 @@
...
+		return true
+	}
+
+	if namespace == "" { // cluster-scoped resources don't have namespace
+		return true
+	}
+
+	allowedNSMap := c.get()
+	c.mu.RLock()
+	defer c.mu.RUnlock()
```
jlpadilla: @coderabbitai Is this lock necessary?

I thought we needed it, but I might be wrong. A different agent continues to flag it as unecessary.

jlpadilla: @smcavey Let's remove this lock.

jlpadilla (approved):
/lgtm

PR #875, opened 2026-05-13
Add metrics for sync timeout and resync track

Sri-glad:
### Description of changes
Add metrics to the search-collector to improve observability of sync failures:

- `search_collector_sync_timeout_total` — Incremented when the collector receives a 504 Gateway Timeout from the aggregated API proxy, indicates the sync request exceeded the 60s timeout.
- `search_collector_resync_total` — Incremented when a diff sync fails and the collector escalates to a complete resync payload.

Large managed clusters (200k+ resources) can exceed the 60 second aggregated API timeout during complete resync, causing the collector to enter an infinite retry loop. The only supported remediation is restarting the search-postgres pod (RH KB https://access.redhat.com/solutions/7134528)

Currently there is no metric to detect this condition, only checking logs can identify affected clusters.

## Summary by CodeRabbit

* **New Features**
  * Added monitoring for sync requests: a new counter tracks total sync HTTP requests, recording response status codes and whether the attempt was a sync or resync. This improves visibility into sync activity and outcomes.

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -21,4 +21,10 @@ var (
 		Name: "search_collector_resources_sent_to_indexer_count",
 		Help: "Total resources sent to indexer after reconciliation",
 	}, []string{"resource_kind"})
+
+	// SyncRequestTotal total number of HTTP requests sent to indexer
+	SyncRequestTotal = promauto.With(PromRegistry).NewCounterVec(prometheus.CounterOpts{
+		Name: "search_collector_sync_requests_total",
+		Help: "Total number of HTTP requests sent",
+	}, []string{"status_code", "sync_type"})
 )
--- a/pkg/send/sender.go
+++ b/pkg/send/sender.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/klog/v2"
 
 	"github.com/stolostron/search-collector/pkg/config"
+	"github.com/stolostron/search-collector/pkg/metrics"
 	"github.com/stolostron/search-collector/pkg/reconciler"
 	tr "github.com/stolostron/search-collector/pkg/transforms"
 )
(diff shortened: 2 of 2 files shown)


Sri-glad:
@coderabbitai review

smcavey:
Thanks for the contribution :+1: Thinking that while we're here trying to capture request types (sync vs resync) and status code 504s we have an opportunity to generalize with something like:

```
diff --git a/pkg/send/sender.go b/pkg/send/sender.go
index eef8f1f..0e71c50 100644
--- a/pkg/send/sender.go
+++ b/pkg/send/sender.go
@@ -211,6 +211,14 @@ func (s *Sender) send(payload Payload, expectedTotalResources int, expectedTotal
                klog.Error("httpClient error: ", err)
                return err
        }
+       //  SyncRequestTotal = promauto.With(PromRegistry).NewCounterVec(prometheus.CounterOpts{
+       //         Name: "search_collector_sync_requests_total",
+       //         Help: "Total number of HTTP requests sent",
+       //  }, []string{"status_code", "sync_type"})
+       // increment countervec with:
+       // metrics.SyncRequestTotal.WithLabelValues(resp.StatusCode, payload.ClearAll).Inc()
+       // label: sync/resync determined from payload.ClearAll
+       // label: responseCode determined from resp.StatusCode
        if resp.StatusCode == http.StatusTooManyRequests {
```

Where we can use a singular new metric to capture total number of requests, whether they're a sync/resync, and capture all response codes. I think that could still accomplish what you're looking to get insight on?

Sri-glad:
@smcavey Indeed, thanks for the suggestion! consolidated into a single CounterVec with status_code and sync_type labels as proposed, captures all HTTP responses and differentiates sync from resync via labels

smcavey:
Looks good deployed in my cluster:
```
# HELP search_collector_sync_requests_total Total number of HTTP requests sent
# TYPE search_collector_sync_requests_total counter
search_collector_sync_requests_total{status_code="200",sync_type="resync"} 1
search_collector_sync_requests_total{status_code="200",sync_type="sync"} 8
```

PR #881, opened 2026-05-19
ACM-33146: Surface CollectorConfig rule errors as status conditions

sahare:
https://redhat.atlassian.net/browse/ACM-33146

## What

When the collector processes a `CollectorConfig` and skips rules due to errors, it now writes those errors back to the CR as a Kubernetes status condition instead of only logging them.

Requires `FEATURE_CONFIGURABLE_COLLECTION=true`.

## Why

Previously the only way to know a rule was silently ignored was to grep pod logs. After this change:

```
oc describe collectorconfig collector-config
Status:
  Conditions:
    Last Transition Time:  2026-05-19T17:51:33Z
    Message:               Rule skipped: only "include" action is supported, found "exclude"; Rule skipped: include action with fields must specify exactly 1 kind, found 2
    Reason:                RulesSkipped
    Status:                False
    Type:                  Applied
```

## Changes

**`pkg/transforms/configurableCollection.go`**
- Accumulates per-rule warning messages into `[]string` instead of just logging them
- New `updateCollectorConfigStatus()` writes the `Applied` condition to the CR via the dynamic client status subresource (best-effort — failure logs a warning, does not abort the collector)
- `lastTransitionTime` follows Kubernetes convention: only updated on True↔False transitions, not on every collector restart

## Tests (34 passing)

Added 13 new tests covering: all warning code paths, multiple warnings in one condition, feature disabled, missing CR, RBAC failure resilience, and `lastTransitionTime` preservation vs update behavior. See `TestStatusCondition_*` in `genericResourceConfig_test.go`.

## E2E (ACM 5.0.0-66)

Deployed a custom `linux/amd64` image and ran 9 live test cases covering every `lastTransitionTime` transition (True→True, True→False, False→False, False→True) and edge cases (empty rules, mixed valid+invalid, no CR, different error same status).  All 9 passed. ✅

## Merge order ⚠️

**This PR must merge AFTER the companion operator PR** ([stolostron/search-v2-operator#726](https://github.com/stolostron/search-v2-operator/pull/726)).

Once the operator PR merges and a new module version is published:
1. Update `go.mod`: `go get github.com/stolostron/search-v2-operator@<new-version>`
2. Run `go mod tidy`
3. Optionally replace the local condition constants with imports from `v1alpha1`
4. Remove draft status and merge

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * CollectorConfig now records an Applied condition (True/False) with reasons: Applied, RulesSkipped, LoadError.
  * Aggregates human-readable skip warnings (truncated with "... and N more") and preserves lastTransitionTime when status is unchanged; status-update failures are logged but do not block merges.

* **Tests**
  * Added tests covering Applied reporting, skipped-rule warnings, truncation, feature-disabled/CR-missing cases, status-update failures, and transition-time behavior.

### DIFF
--- a/pkg/transforms/configurableCollection.go
+++ b/pkg/transforms/configurableCollection.go
@@ -238,17 +238,22 @@ func updateCollectorConfigStatus(dynamicClient dynamic.Interface, namespace stri
 	}
 
 	// Preserve lastTransitionTime if the condition status hasn't changed.
-	// Only update it when True→False or False→True to follow Kubernetes conventions.
+	// Search by type (not by index) so other conditions don't affect the lookup.
+	// Only update the timestamp on True↔False transitions per Kubernetes conventions.
 	lastTransitionTime := metav1.Now().UTC().Format("2006-01-02T15:04:05Z")
-	if existing, ok := configObj.Object["status"].(map[string]interface{}); ok {
-		if conditions, ok := existing["conditions"].([]interface{}); ok && len(conditions) > 0 {
-			if cond, ok := conditions[0].(map[string]interface{}); ok {
-				if cond["type"] == collectorConfigConditionApplied &&
-					cond["status"] == string(conditionStatus) {
-					if t, ok := cond["lastTransitionTime"].(string); ok && t != "" {
-						lastTransitionTime = t
-					}
+	existingStatus, _ := configObj.Object["status"].(map[string]interface{})
+	if existingStatus == nil {
+		existingStatus = map[string]interface{}{}
+	}
+	if conditions, ok := existingStatus["conditions"].([]interface{}); ok {
+		for _, c := range conditions {
+			if cond, ok := c.(map[string]interface{}); ok &&
+				cond["type"] == collectorConfigConditionApplied &&
+				cond["status"] == string(conditionStatus) {
+				if t, ok := cond["lastTransitionTime"].(string); ok && t != "" {
+					lastTransitionTime = t
 				}
+				break
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


sahare:
@coderabbitai review

pkg/transforms/configurableCollection.go
```diff
@@ -168,9 +211,86 @@ func loadAndMergeConfigurableCollectionWithClient(dynamicClient dynamic.Interfac
...
+// the last time the condition was evaluated.
+// This is best-effort: failures are logged but do not abort the collector.
+func updateCollectorConfigStatus(dynamicClient dynamic.Interface, namespace string,
+	configObj *unstructured.Unstructured, warnings []string, reason string) {
+
+	conditionStatus := metav1.ConditionTrue
+	message := "Configuration applied successfully."
+	if len(warnings) > 0 {
+		conditionStatus = metav1.ConditionFalse
+		message = strings.Join(warnings, "; ")
```
jlpadilla: I wonder how this scales if there's a lot of warnings.
Maybe we should truncate at 3 warnings and add something like "5 more..." to indicate it was truncated?

pkg/transforms/genericResourceConfig_test.go
```diff
@@ -1006,6 +1010,665 @@ func TestLoadAndMergeConfigurableCollection_FieldCollisionWithSuffix(t *testing.
...
+
+	assert.Equal(t, "False", cond["status"])
+	assert.Equal(t, "RulesSkipped", cond["reason"])
+
+	msg, _ := cond["message"].(string)
+	// All 3 warnings should be in the single message separated by "; "
+	assert.True(t, strings.Contains(msg, "exclude"), "Message should mention the exclude action, got: %s", msg)
+	assert.True(t, strings.Contains(msg, "requires at least one field"), "Message should mention missing fields, got: %s", msg)
+	assert.True(t, strings.Contains(msg, "exactly 1 kind"), "Message should mention multiple kinds, got: %s", msg)
+	assert.True(t, strings.Contains(msg, "; "), "Multiple warnings should be separated by '; ', got: %s", msg)
```
smcavey, line 1420: Could we be a little more verbose and add:
```
assert.True(t, strings.Contains(msg, "; ")......
assert.True(t, strings.Contains(msg, "only \"include\" action is supported")...... // Rule 1
assert.True(t, strings.Contains(msg, "include action requires at least one field")...... // Rule 2
assert.True(t, strings.Contains(msg, "include action with fields must specify exactly 1 kind")...... // Rule 3
```

jlpadilla:
Change in search-v2-operator has been propagated to multiclusterhub-operator

https://github.com/stolostron/multiclusterhub-operator/pull/4176

jlpadilla (approved):
/lgtm

PR #882, opened 2026-05-28
ACM-30764: Store boolean resource properties as strings

sahare:
https://redhat.atlassian.net/browse/ACM-30764

## Problem

Boolean fields like `allowVolumeExpansion: true` on StorageClass returned **zero search results** even when matching resources existed.

The search API queries JSONB fields as strings (`WHERE data->'field' ? 'true'`), but these fields were stored as JSON booleans — not strings — so the `?` operator never matched them.

**Affected fields (9 total):**
`disabled` (Policy), `suspend` (CronJob), `background`/`admission` (Kyverno), `allowVolumeExpansion` (StorageClass), `allowAutoConverge`/`allowPostCopy` (MigrationPolicy), `readyToUse` (VirtualMachineSnapshot), `complete` (VirtualMachineRestore)

## Fix

Two-part approach (aligns with closed PR #839 by @smcavey):

**1. Centralized (defensive):** In `applyDefaultTransformConfig` (`common.go`), any boolean value extracted via JSONPath is automatically converted to its string representation with `strconv.FormatBool()`. Future boolean fields are handled automatically.

**2. Targeted:** Explicit `strconv.FormatBool()` conversions in typed builders that hardcode boolean assignments: `policy.go` (`disabled`), `cronjob.go` (`suspend`), `kyverno.go` (`background`, `admission`).

## Tests

- Updated all existing assertions that expected `bool` values to expect `string` values
- Added 21 new tests covering every affected field × `true`/`false` combinations, nil pointer handling, and a dedicated **regression test** (`TestBooleanFieldsNotStoredAsRawBool_RegressionPrevention`) that explicitly fails if any field reverts to storing a raw boolean

## ⚠️ Breaking change / Errata note

Fields previously stored as JSON booleans (`true`/`false`) are now stored as strings (`"true"`/`"false"`). This is intentional — the API already treated them as strings for queries.

**Impact:** Any integration reading raw boolean values from the search index directly will now receive strings instead. Console queries and standard API usage are unaffected (actually improved — queries now return results).

This change is appropriate for the 5.0 major version boundary.

## E2E Test Results
Tested on ACM 5.0.0-83.
**Before fix:** `kind:StorageClass allowVolumeExpansion:true` returned 0 results. PostgreSQL stored the field as a JSON boolean so the API's string-based query couldn't match it.
**After fix:** 
- `jsonb_typeof(data->'allowVolumeExpansion')` → `string` ✅  
- `kind:StorageClass allowVolumeExpansion:true` → `count: 2` ✅  
- `kind:Policy disabled:true` → `count: 1` ✅  

## Summary by CodeRabbit

* **Bug Fixes**
  * Boolean-type resource properties are now consistently stored as string values ("true"/"false") across multiple resources (suspend, policy flags, placement, Kyverno settings, etc.), improving serialization consistency.

* **Tests**
  * Added and updated unit tests to assert boolean fields are serialized as strings and to prevent regressions across affected resource transforms.

### DIFF
--- a/pkg/transforms/genericResource_test.go
+++ b/pkg/transforms/genericResource_test.go
@@ -3,6 +3,7 @@
 package transforms
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/stolostron/search-collector/pkg/config"
@@ -166,7 +167,7 @@ func Test_genericResourceFromConfigVM(t *testing.T) {
 	AssertEqual("runStrategy", node.Properties["runStrategy"], "always", t)
 	AssertEqual("status", node.Properties["status"], "Running", t)
 	AssertEqual("workload", node.Properties["workload"], "server", t)
-	AssertEqual("_specRunning", node.Properties["_specRunning"], true, t)
+	AssertEqual("_specRunning", node.Properties["_specRunning"], "true", t)
 	AssertEqual("_specRunStrategy", node.Properties["_specRunStrategy"], "always", t)
 }
 
@@ -244,7 +245,7 @@ func Test_genericResourceFromConfigVMSnapshot(t *testing.T) {
 	AssertEqual("ready", node.Properties["ready"], "True", t)
 	AssertEqual("_conditionReadyReason", node.Properties["_conditionReadyReason"], "Operation complete", t)
 	AssertEqual("phase", node.Properties["phase"], "Succeeded", t)
-	AssertEqual("readyToUse", node.Properties["readyToUse"], true, t)
+	AssertEqual("readyToUse", node.Properties["readyToUse"], "true", t)
 	AssertEqual("sourceName", node.Properties["sourceName"], "centos7-gray-owl-35", t)
 	AssertEqual("sourceKind", node.Properties["sourceKind"], "VirtualMachine", t)
 	AssertDeepEqual("indications", node.Properties["indications"], []interface{}{"Online", "NoGuestAgent"}, t)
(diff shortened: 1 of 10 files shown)


sahare:
@coderabbitai review

smcavey:
found a couple more
- https://github.com/stolostron/search-collector/blob/9c8b22a920d07c125fd9e49f196694427147a4dd/pkg/transforms/subscription.go#L50-L53
- https://github.com/stolostron/search-collector/blob/9c8b22a920d07c125fd9e49f196694427147a4dd/pkg/transforms/policy.go#L132-L133
-

pkg/transforms/kyverno.go
```diff
@@ -52,8 +53,8 @@ func KyvernoPolicyResourceBuilder(p *unstructured.Unstructured) *KyvernoPolicyRe
 		}
 	}
 
-	node.Properties["background"] = background
-	node.Properties["admission"] = admission
+	node.Properties["background"] = strconv.FormatBool(background)
+	node.Properties["admission"] = strconv.FormatBool(admission)
```
jlpadilla, line 57: This may break the GRC integration. @JustinKuli @yiraeChristineKim 

We are making this change to consistently store all boolean values as strings. We are doing it for 5.0 because it's a potentially breaking change.

pkg/transforms/genericResource_test.go
```diff
@@ -3,6 +3,7 @@
 package transforms
 
 import (
+	"fmt"
```
smcavey: linter is failing for this not being used anymore after removing the other tests:
```
 CGO_ENABLED=1 GOGC=25 golangci-lint run --timeout=3m
pkg/transforms/appdeployable.go:1: : # github.com/stolostron/search-collector/pkg/transforms [github.com/stolostron/search-collector/pkg/transforms.test]
pkg/transforms/genericResource_test.go:6:2: "fmt" imported and not used (typecheck)
/*
1 issues:
* typecheck: 1
```

pkg/transforms/kyverno.go
JustinKuli: Thanks for the heads up! I'll keep an eye on it

PR #885, opened 2026-06-09
ACM-32738 - configurable collection - additionalPrinterColumns

smcavey:
### Related Issue

https://issues.redhat.com/browse/ACM-32738

### Description of changes
- Added ability to configure collection of additional printer columns via configurable collection
- Pending CRD and webhook updates

## Summary by CodeRabbit

* **New Features**
  * Additional printer columns can have per-column priorities; collection is gated by configurable priority thresholds (including wildcard defaults) so only columns within the threshold are included.

* **Tests**
  * Added tests covering priority parsing, threshold-based filtering, wildcard/default threshold behavior, conflict-resolution (max-wins), and combined collection of regular properties plus printer columns.

### DIFF
--- a/pkg/transforms/common.go
+++ b/pkg/transforms/common.go
@@ -706,10 +706,11 @@ func applyDefaultTransformConfig(node Node, r *unstructured.Unstructured, additi
 			continue
 		}
 
-		// Skip additionalPrinterColumns when not configured or below threshold
+		// Skip additionalPrinterColumns when not configured or above threshold.
+		// A threshold of N means collect columns with priority 0 through N.
 		if prop.Priority != nil {
 			if transformConfig.additionalPrinterColumnsPriority == nil ||
-				*prop.Priority < *transformConfig.additionalPrinterColumnsPriority {
+				*prop.Priority > *transformConfig.additionalPrinterColumnsPriority {
 				continue
 			}
 		}
--- a/pkg/transforms/common.go
+++ b/pkg/transforms/common.go
@@ -666,12 +666,20 @@ func applyDefaultTransformConfig(node Node, r *unstructured.Unstructured, additi
 		}
 	}
 
-	// Currently, only pull in the additionalPrinterColumns listed in the CRD if it's a Gatekeeper
-	// constraint or globally enabled.
-	if !found && (config.Cfg.CollectCRDPrinterColumns || group == "constraints.gatekeeper.sh") {
-		transformConfig = ResourceConfig{properties: additionalColumns}
-	} else if !found {
-		return node
+	// Pull in additionalPrinterColumns when globally enabled, gatekeeper constraint,
+	// or when a wildcard config has only additionalPrinterColumnsPriority set.
+	if !found {
+		useColumns := config.Cfg.CollectCRDPrinterColumns || group == "constraints.gatekeeper.sh" ||
+			(wildcardFound && wildcardConfig.additionalPrinterColumnsPriority != nil)
+		if useColumns {
+			transformConfig = ResourceConfig{properties: additionalColumns}
+			// Inherit the priority threshold from the wildcard config.
+			if wildcardFound && wildcardConfig.additionalPrinterColumnsPriority != nil {
+				transformConfig.additionalPrinterColumnsPriority = wildcardConfig.additionalPrinterColumnsPriority
+			}
+		} else {
+			return node
+		}
 	}
 
 	for _, prop := range transformConfig.properties {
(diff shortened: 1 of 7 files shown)


pkg/transforms/common.go
```diff
@@ -696,6 +704,14 @@ func applyDefaultTransformConfig(node Node, r *unstructured.Unstructured, additi
 			continue
 		}
 
+		// Skip additionalPrinterColumns when not configured or below threshold
+		if prop.Priority != nil {
+			if transformConfig.additionalPrinterColumnsPriority == nil ||
+				*prop.Priority < *transformConfig.additionalPrinterColumnsPriority {
+				continue
+			}
+		}
```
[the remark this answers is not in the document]
smcavey: Here's some context about the distribution of priorities in a pretty plain ACM install:
695 additionalPrinterColumns:

- 632 have no priority
- 60 have priority 1
- 3 have priority 10

I find it easier to specify priority 0 to collect all additional printer columns, considering most are priority 0, than it is to find the largest int priority and use that as my threshold to collect all additional printer columns.

pkg/transforms/common_test.go
```diff
@@ -295,6 +295,165 @@ func Test_genericResourceFromConfigMapNoLabelsToMatchMatchLabel(t *testing.T) {
...
+			"kind":       "Fake",
+			"metadata":   map[string]interface{}{"name": "test", "uid": "uid-1", "creationTimestamp": "2024-01-01T00:00:00Z"},
+			"status":     map[string]interface{}{"phase": "Running", "ready": "True"},
+		},
+	}
+
+	node := GenericResourceBuilder(&r).BuildNode()
+	assert.Nil(t, node.Properties["lowPriority"], "Priority-0 column should be skipped when threshold is 1")
+	assert.Equal(t, "True", node.Properties["highPriority"], "Priority-1 column should be collected when threshold is 1")
+}
```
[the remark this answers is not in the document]
smcavey: See https://github.com/stolostron/search-collector/pull/885#discussion_r3380611460

pkg/transforms/configurableCollection.go
```diff
@@ -392,6 +396,37 @@ func mergeCollectAnnotations(apiGroups, kinds []string) {
...
+			resourceConfig, exists := mergedTransformConfig[resourceKey]
+			if !exists {
+				resourceConfig = ResourceConfig{
+					properties: []ExtractProperty{},
+				}
+			}
+			// Take the max priority — higher values are more permissive (collect more columns).
+			// This prevents a later rule from accidentally narrowing an earlier rule's threshold.
+			if resourceConfig.additionalPrinterColumnsPriority == nil || priority > *resourceConfig.additionalPrinterColumnsPriority {
+				resourceConfig.additionalPrinterColumnsPriority = &priority
```
jlpadilla, line 422: **🤖 AI Agent Review**
Storing `&priority` where `priority` is a function parameter is non-idiomatic — while Go's escape analysis will heap-allocate the variable here (because the address escapes into the map), this is an implementation detail of the compiler rather than a language guarantee. The pattern used in `deepCopyTransformConfig` (lines 461–464) is the clearer approach:

```suggestion
			p := priority
				resourceConfig.additionalPrinterColumnsPriority = &p
```

pkg/transforms/common.go
```diff
@@ -666,12 +666,20 @@ func applyDefaultTransformConfig(node Node, r *unstructured.Unstructured, additi
...
 
-	// Currently, only pull in the additionalPrinterColumns listed in the CRD if it's a Gatekeeper
-	// constraint or globally enabled.
-	if !found && (config.Cfg.CollectCRDPrinterColumns || group == "constraints.gatekeeper.sh") {
-		transformConfig = ResourceConfig{properties: additionalColumns}
-	} else if !found {
-		return node
+	// Pull in additionalPrinterColumns when globally enabled, gatekeeper constraint,
+	// or when a wildcard config has only additionalPrinterColumnsPriority set.
+	if !found {
```
jlpadilla, line 671: **🤖 AI Agent Review**
The wildcard `additionalPrinterColumnsPriority` threshold is only inherited inside the `!found` branch, meaning it only applies to resources that have **no specific entry** in `mergedTransformConfig`. If a resource already has a specific `ResourceConfig` (e.g., because another rule added fields or conditions for it), this wildcard printer-column threshold is silently ignored — the `additionalColumns` (CRD printer columns) are also never appended to the specific config in that branch.

This is a design gap worth documenting even if not fixed in this PR, so operators know that wildcard `collectAdditionalPrinterColumnsPriority` rules only take effect for resources without any other specific configuration.

jlpadilla (commented):
**🤖 AI Agent Review**
Good feature — the priority comparison, max-wins merge, and CRD parsing are all correct. Tests are thorough and SonarCloud is happy.

A few things worth looking at before merge — see inline comments. The main one is the wildcard threshold gap (line 671 in common.go): if a resource has any specific config entry, the wildcard printer-column threshold is silently ignored. Worth at least documenting as a known limitation.

Also left a note on the `&priority` parameter alias in `mergeCollectPrinterColumns` (safe in practice but non-idiomatic). And separately — `deepCopyTransformConfig` does a shallow copy of `ExtractProperty.Priority` pointers (line 454, not in the diff since it predates this PR). That is latent today but could bite if CRD columns are ever persisted into `mergedTransformConfig`.

pkg/transforms/common.go
jlpadilla: This is an interesting finding, but we can merge without this.

jlpadilla (approved):
/lgtm

pkg/transforms/configurableCollection.go
jlpadilla: very minor, approved as-is.

PR #888, opened 2026-06-18
ACM-35522: Implement exclude action for CollectorConfig collection rules

sahare:
## Summary

Implements `action: exclude` in `CollectorConfig` collection rules (ACM-35522), allowing users to stop the search-collector from indexing specific resource types.

**Companion PR:** stolostron/search-v2-operator#740 (webhook validation + merge protection)

## Changes

**`pkg/transforms/configurableCollection.go`**

- `excludeRules []excludeRule` — ordered slice replacing the old `excludedResources map[string]struct{}` (compound-key map). Each entry is `{apiGroups, kinds, action}` matching the same cartesian wildcard logic as `isResourceMatchingList` in `supportedResources.go`, aligning the exclude implementation with the existing Allow/Deny feature.
- `appendExcludeRule()` — replaces `mergeExcludeRules` + `unmergeExcludeRules`. Both exclude and valid include rules are appended to the list.
- `IsResourceExcluded()` — walks the ordered list; last matching rule determines the outcome (last-match-wins). An ActionInclude entry cancels any prior ActionExclude for the same resource.
- `matchesExcludeRule()` — shared cartesian `(g=="*" || g==group) && (k=="*" || k==kind)` logic.

**Wildcard-vs-specific resolved:**
`exclude "*.*"` + `include "Deployment.apps"` → Deployments are collected, all other resources excluded. This is the primary migration use case for users moving from the AllowedResources ConfigMap feature.

**`pkg/informer/informer.go`**

- `IsResourceExcluded()` gate at all 3 event entry points (list sync, ADDED, MODIFIED).

## Design decisions

- Exclude is enforced at the **informer layer** so resources are stopped before any transform work.
- Integration team protection is handled in the **operator** (merge time), not the collector.
- "Last entry wins" with proper wildcard resolution: a specific include overrides a prior wildcard exclude.
- `matchesExcludeRule` uses the same cartesian logic as `isResourceMatchingList` for consistency with Allow/Deny.

## Known design behavior

**Stale records after adding an exclude rule:** Previously indexed resources remain in the search DB until deleted from the cluster. This is an inherent consequence of informer-layer enforcement.

## Test plan

- `go test ./pkg/transforms/` — 12 exclude tests pass including `TestExclude_WildcardOverriddenBySpecificInclude`
- `go test ./...` — all tests pass, `golangci-lint` 0 issues
- E2E on cluster pending (see Jira ACM-35522 for full test matrix)

### DIFF
--- a/pkg/informer/supportedResources.go
+++ b/pkg/informer/supportedResources.go
@@ -142,6 +142,18 @@ func SupportedResources(discoveryClient discovery.DiscoveryClient) (map[schema.G
 				continue // Skip the resource before starting the informer
 			}
 
+			// Skip resources excluded via CollectorConfig exclude rules.
+			// Normalize the group to empty string for core-group resources ("v1" → "")
+			// to match the CollectorConfig apiGroups convention.
+			apiGroup := group
+			if len(groupVersion) == 1 {
+				apiGroup = ""
+			}
+			if tr.IsResourceExcluded(apiGroup, apiResource.Kind) {
+				klog.V(3).Infof("Skipping excluded resource [group: '%s' kind: %s]. Matched CollectorConfig exclude rule.", apiGroup, apiResource.Kind)
+				continue
+			}
+
 			// add non-namespaced resource to NonNSResourceMap
 			if !apiResource.Namespaced {
 				tr.NonNSResMapMutex.Lock()
--- a/pkg/informer/supportedResources.go
+++ b/pkg/informer/supportedResources.go
@@ -150,7 +150,8 @@ func SupportedResources(discoveryClient discovery.DiscoveryClient) (map[schema.G
 				apiGroup = ""
 			}
 			if tr.IsResourceExcluded(apiGroup, apiResource.Kind) {
-				klog.V(3).Infof("Skipping excluded resource [group: '%s' kind: %s]. Matched CollectorConfig exclude rule.", apiGroup, apiResource.Kind)
+				klog.V(3).Infof("Skipping excluded resource [group: '%s' kind: %s]."+
+					" Matched CollectorConfig exclude rule.", apiGroup, apiResource.Kind)
 				continue
 			}
 
--- a/pkg/informer/supportedResources.go
+++ b/pkg/informer/supportedResources.go
@@ -151,7 +151,7 @@ func SupportedResources(discoveryClient discovery.DiscoveryClient) (map[schema.G
 			}
 			if tr.IsResourceExcluded(apiGroup, apiResource.Kind) {
 				klog.V(3).Infof("Skipping excluded resource [group: '%s' kind: %s]."+
-					" Matched CollectorConfig exclude rule.", apiGroup, apiResource.Kind)
+					" Matched CollectorConfig exclude rule.", apiGroup, apiResource.Name)
 				continue
 			}
 
--- a/pkg/transforms/configurableCollection.go
+++ b/pkg/transforms/configurableCollection.go
@@ -20,13 +20,19 @@ import (
 // Wildcard entries like "*" (core group) or "*.apps" are used for apigroup-wide collectConditions.
 var mergedTransformConfig map[string]ResourceConfig
 
+// excludedResources holds the set of resource keys that should not be collected.
+// Keys follow the same scheme as mergedTransformConfig: "Kind.apiGroup" for specific kinds,
+// "*" for all core-group resources, "*.apps" for all resources in the apps apiGroup.
+// Populated by mergeExcludeRules during config load; checked at the informer layer.
+var excludedResources map[string]struct{}
+
 // LoadAndMergeConfigurableCollection loads the CollectorConfig resource from the cluster and merges it with defaultTransformConfig.
 // The merged result is stored in mergedTransformConfig.
 func LoadAndMergeConfigurableCollection() {
 	if !config.Cfg.FeatureConfigurableCollection {
 		klog.Info("Configurable collection feature is disabled, skipping custom config load")
-		// Initialize mergedTransformConfig to a copy of defaultTransformConfig
 		mergedTransformConfig = deepCopyTransformConfig(defaultTransformConfig)
+		excludedResources = map[string]struct{}{}
 		return
 	}
 
@@ -54,8 +60,9 @@ var collectorConfigGVR = schema.GroupVersionResource{
 
 // loadAndMergeConfigurableCollectionWithClient is a helper function that accepts a dynamic client for testability.
 func loadAndMergeConfigurableCollectionWithClient(dynamicClient dynamic.Interface) {
-	// Start with a deep copy of defaultTransformConfig
+	// Start with a deep copy of defaultTransformConfig and a fresh exclude set.
 	mergedTransformConfig = deepCopyTransformConfig(defaultTransformConfig)
+	excludedResources = map[string]struct{}{}
 
 	namespace := config.Cfg.PodNamespace
 
@@ -99,13 +106,16 @@ func loadAndMergeConfigurableCollectionWithClient(dynamicClient dynamic.Interfac
 		// Get field suffix for this rule (defaults to empty string)
 		fieldSuffix := rule.FieldSuffix
 
-		// FUTURE: Only Include actions are currently supported
-		if rule.Action != v1alpha1.ActionInclude {
-			msg := fmt.Sprintf("Rule skipped: only \"include\" action is supported, found %q", rule.Action)
-			klog.Warningf("Skipping collection rule. Only \"include\" action supported at this time: %s", rule.Action)
-			warnings = append(warnings, msg)
+		if rule.Action == v1alpha1.ActionExclude {
+			mergeExcludeRules(rule.ResourceSelector.APIGroups, rule.ResourceSelector.Kinds)
 			continue
 		}
+
+		// "Last entry wins": an include rule cancels any prior exclude for the same resource.
+		// Uses the same key scheme as mergeExcludeRules so exact keys are matched and removed.
+		// Note: wildcard-vs-specific mismatches (e.g. exclude "*.*" followed by include
+		// "Deployment.apps") are not resolved — those are documented as a known limitation.
+		unmergeExcludeRules(rule.ResourceSelector.APIGroups, rule.ResourceSelector.Kinds)
 
 		hasFields := len(rule.Fields) > 0
 		hasCollectConditions := rule.CollectConditions != nil && *rule.CollectConditions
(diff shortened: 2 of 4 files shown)


pkg/transforms/configurableCollection.go
```diff
@@ -99,14 +106,17 @@ func loadAndMergeConfigurableCollectionWithClient(dynamicClient dynamic.Interfac
...
+		if rule.Action == v1alpha1.ActionExclude {
+			mergeExcludeRules(rule.ResourceSelector.APIGroups, rule.ResourceSelector.Kinds)
 			continue
 		}
 
+		// "Last entry wins": an include rule cancels any prior exclude for the same resource.
+		// Uses the same key scheme as mergeExcludeRules so exact keys are matched and removed.
+		// Note: wildcard-vs-specific mismatches (e.g. exclude "*.*" followed by include
+		// "Deployment.apps") are not resolved — those are documented as a known limitation.
+		unmergeExcludeRules(rule.ResourceSelector.APIGroups, rule.ResourceSelector.Kinds)
```
sahare: **Bug: `unmergeExcludeRules` fires before the actionable-content guard.**

`unmergeExcludeRules` is called here for every non-exclude rule, but the check that determines whether the include rule is actually valid (has fields/collectConditions/collectAnnotations) comes later at line 125. An empty/malformed include rule that gets a warning and a `continue` will still have already removed any prior exclude for the same resource.

Reproducer:
```yaml
collectionRules:
  - action: exclude
    resourceSelector: {apiGroups: ["coordination.k8s.io"], kinds: ["Lease"]}
  - action: include
    resourceSelector: {apiGroups: ["coordination.k8s.io"], kinds: ["Lease"]}
    # no fields, no collectConditions — skipped with a warning
```
Result: `Lease` is **not** excluded. The user sees `Applied=False / RulesSkipped` (warning about the include) but the exclusion was silently cancelled — counterintuitive.

Fix: move `unmergeExcludeRules` to *after* the `!hasFields && !hasCollectConditions && !hasCollectAnnotations` guard so that only actually-applied include rules cancel prior excludes.

pkg/transforms/configurableCollection_test.go
```diff
@@ -2720,3 +2784,264 @@ func TestLoadAndMergeConfigurableCollection_CollectAnnotationsOnly(t *testing.T)
...
+	assert.True(t, IsResourceExcluded("coordination.k8s.io", "Lease"),
+		"Lease.coordination.k8s.io should be excluded")
+	assert.False(t, IsResourceExcluded("coordination.k8s.io", "LeaderElection"),
+		"Other kinds in group should not be excluded")
+	assert.False(t, IsResourceExcluded("", "Lease"),
+		"Core-group Lease should not be excluded")
+}
+
+// Wildcard kind excludes all kinds in a group.
+func TestExclude_WildcardKind(t *testing.T) {
```
sahare, line 3225: The four existing wildcard tests cover exact match, `*.group`, `*.*`, and empty apiGroup. Check #3 in `IsResourceExcluded` — the `"Kind.*"` (group wildcard) path — is never exercised. Worth adding a `TestExclude_GroupWildcard` case:

```go
// TestExclude_GroupWildcard verifies that exclude apiGroups=["*"], kinds=["Lease"]
// stores "Lease.*" and matches Lease in any non-core apiGroup.
func TestExclude_GroupWildcard(t *testing.T) {
    teardown, _ := setupExcludeTest(t)
    defer teardown()

    cfg := makeExcludeConfig("test-namespace", "*", "Lease")
    fakeClient := fake.NewSimpleDynamicClient(runtime.NewScheme(), cfg)
    loadAndMergeConfigurableCollectionWithClient(fakeClient)

    assert.True(t, IsResourceExcluded("coordination.k8s.io", "Lease"),
        "Lease.coordination.k8s.io should match Lease.*")
    assert.True(t, IsResourceExcluded("someother.io", "Lease"),
        "Lease.someother.io should also match Lease.*")
    assert.False(t, IsResourceExcluded("apps", "Deployment"),
        "Deployment should not be excluded")
}
```

pkg/transforms/configurableCollection_test.go
```diff
@@ -103,12 +103,15 @@ func TestLoadAndMergeConfigurableCollection_ValidConfig(t *testing.T) {
 }
 
 // FUTURE: this should eventually appropriately include when search-collector-config merged
```
sahare: This `// FUTURE:` comment is left over from when `exclude` was not yet implemented. Now that it is fully supported, the comment is stale and potentially confusing. Can drop the first line.

```suggestion
// Exclude rules do not add entries to mergedTransformConfig — they populate excludedResources instead.
```

pkg/transforms/configurableCollection.go
```diff
@@ -20,13 +20,19 @@ import (
 // Wildcard entries like "*" (core group) or "*.apps" are used for apigroup-wide collectConditions.
 var mergedTransformConfig map[string]ResourceConfig
 
+// excludedResources holds the set of resource keys that should not be collected.
+// Keys follow the same scheme as mergedTransformConfig: "Kind.apiGroup" for specific kinds,
+// "*" for all core-group resources, "*.apps" for all resources in the apps apiGroup.
+// Populated by mergeExcludeRules during config load; checked at the informer layer.
+var excludedResources map[string]struct{}
```
sahare: **Thread safety note for ACM-20047.** `excludedResources` is read by every informer goroutine (one per GVR) on each ADDED/MODIFIED/LIST event. Currently safe because `LoadAndMergeConfigurableCollection` is called once at startup before any informers are started, so there is no concurrent write. However, `nsFilterCache` (the analogous namespace-filter state) uses a `sync.RWMutex` for exactly this reason — it can be refreshed mid-run.

When ACM-20047 (dynamic config reload) is implemented, `excludedResources` will be written while informers are running, creating a data race. A short comment here (or a `sync.RWMutex` now) would prevent the oversight:

```suggestion
// excludedResources holds the set of resource keys that should not be collected.
// Keys follow the same scheme as mergedTransformConfig: "Kind.apiGroup" for specific kinds,
// "*" for all core-group resources, "*.apps" for all resources in the apps apiGroup.
// Populated by mergeExcludeRules during config load; checked at the informer layer.
// NOTE: currently written once at startup before informers run (no concurrent write).
// When ACM-20047 adds dynamic reload, this will need a sync.RWMutex like nsFilterCache.
var excludedResources map[string]struct{}
```

sahare (commented):
Solid implementation overall — the informer-layer enforcement, 4-level wildcard lookup, and last-entry-wins semantics are all well-designed and well-tested. One behavioral bug to address before merging: a skipped include rule (no fields/conditions) silently cancels a prior exclude, which is counterintuitive. See inline comments for the fix + a couple of smaller items.

pkg/transforms/configurableCollection_test.go
sahare: The `// FUTURE:` line was not removed by the fix commit — only the second comment line was reworded. The intent was to drop this stale line since `exclude` is now fully implemented.

```suggestion
// Exclude rules populate excludedResources (not mergedTransformConfig).
```

pkg/transforms/configurableCollection.go
sahare: Thread safety note still missing — this was issue #4 from the first pass. `nsFilterCache` has an analogous comment, and this will need a `sync.RWMutex` when ACM-20047 (dynamic config reload) lands.

```suggestion
// NOTE: currently written once at startup before informers run (no concurrent write).
// When ACM-20047 adds dynamic reload, this will need a sync.RWMutex like nsFilterCache.
var excludedResources map[string]struct{}
```

sahare (commented):
Second pass — the critical fix landed correctly. Bug fix is sound, regression test covers the exact scenario, and `IsResourceExcluded` check #3 now correctly matches core-group resources. Two small nits from the first pass remain (stale FUTURE comment, thread-safety note) — see inline comments. Nothing blocking merge.

pkg/transforms/configurableCollection.go
sahare: Fixed in commit db75cc1 — unmergeExcludeRules is now called after the actionable-content guard (hasFields / hasCollectConditions / hasCollectAnnotations check), so only valid include rules can cancel a prior exclude. Regression test TestExclude_InvalidIncludeDoesNotCancelExclude covers this exact scenario.

pkg/transforms/configurableCollection_test.go
sahare: Added TestExclude_GroupWildcard in commit db75cc1 — covers exact match, non-core group, and negative case. Also fixed IsResourceExcluded check #3 to apply even when group is empty so apiGroups:[*] correctly excludes core-group resources.

pkg/transforms/configurableCollection_test.go
sahare: Fixed in commit 6b668fd — the FUTURE comment line was removed; the function comment now reads 'Exclude rules populate excludedResources (not mergedTransformConfig).'

pkg/transforms/configurableCollection.go
sahare: Added in commit 6b668fd — thread safety note added to the excludedResources declaration explaining the current safe-because-startup guarantee and flagging that ACM-20047 will require a sync.RWMutex like nsFilterCache.

pkg/transforms/configurableCollection_test.go
sahare: Fixed in commit 6b668fd — the stale FUTURE line was removed in that commit.

pkg/transforms/configurableCollection.go
sahare: Added in commit 6b668fd — thread safety note is now present on the excludedResources declaration.

sahare:
@coderabbitai review

pkg/transforms/configurableCollection.go
```diff
@@ -99,26 +111,30 @@ func loadAndMergeConfigurableCollectionWithClient(dynamicClient dynamic.Interfac
...
 			msg := "Rule skipped: include action requires at least one field, collectConditions, or collectAnnotations"
 			klog.Warning("Skipping collection rule. Include action without fields, collectConditions, or collectAnnotations specified.")
 			warnings = append(warnings, msg)
 			continue
 		}
 
+		// "Last entry wins": a valid include rule cancels any prior exclude for the same resource.
+		// Uses the same key scheme as mergeExcludeRules so exact keys are matched and removed.
+		// Note: wildcard-vs-specific mismatches (e.g. exclude "*.*" followed by include
+		// "Deployment.apps") are not resolved — those are documented as a known limitation.
```
jlpadilla: This is a significant limitation. We currently support this in the existing Allow/Deny fearute which we plan to migrate to the new CollectorConfig.

jlpadilla: Another reason this is significant is because our default we collect everything. A user that want total control of what is collected needs to start by denying `*.*` and then only add what they want to collect.

pkg/transforms/configurableCollection.go
```diff
@@ -20,13 +20,24 @@ import (
...
+// excludedResources holds the set of resource keys that should not be collected.
+// Keys follow the same scheme as mergedTransformConfig: "Kind.apiGroup" for specific kinds,
+// "*" for all core-group resources, "*.apps" for all resources in the apps apiGroup.
+// Populated by mergeExcludeRules during config load; checked at the informer layer.
+//
+// Thread safety: currently written once at startup (before informers start) and read
+// concurrently by informer goroutines. Safe for now because reads begin only after
+// LoadAndMergeConfigurableCollection returns. When ACM-20047 adds dynamic reload,
+// this will need a sync.RWMutex — see nsFilterCache for the expected pattern.
+var excludedResources map[string]struct{}
```
jlpadilla: I think this should be:
```suggestion
var excludedResources map[string]string[]
```

Then the data will look like this:
```
{ 
"api-group-1": ["*"],
"api-group-2": ["kind-a", "kind-b"],
}

Comment (jlpadilla) [reply pkg/transforms/configurableCollection.go]: Actually, it should be 
```
var excludedResources map[string]Resource
```
https://github.com/stolostron/search-collector/blob/main/pkg/informer/supportedResources.go#L19-L22

**We have an existing implementation for Allow/Deny resources. Check the file above.
The goal is to replace that with the new CollectorConfig.**

File: pkg/transforms/configurableCollection.go [hunk 7]
```diff
@@ -342,6 +373,54 @@ func updateCollectorConfigStatus(dynamicClient dynamic.Interface, namespace stri
...
+}
+
+// IsResourceExcluded reports whether a resource should be excluded from collection.
+// Evaluates the ordered excludeRules list with "last matching rule wins" semantics:
+//   - An ActionExclude rule that matches marks the resource as excluded.
+//   - An ActionInclude rule that matches cancels any prior exclude.
+//
+// This resolves wildcard-vs-specific correctly: exclude "*.*" followed by
+// include "Deployment.apps" results in Deployments being collected (not excluded).
+func IsResourceExcluded(group, kind string) bool {
```
jlpadilla, line 414: This check runs when the informer receives an event. This means that an informer gets initialized, but all the events are discarded later. This is inefficient and it would be better to not initialize the informer in the first place.

Here's the logic for the existing Allow/Deny implementation.
https://github.com/stolostron/search-collector/blob/main/pkg/informer/supportedResources.go#L141-L143

pkg/informer/supportedResources.go
```diff
@@ -142,6 +142,19 @@ func SupportedResources(discoveryClient discovery.DiscoveryClient) (map[schema.G
...
+			// Skip resources excluded via CollectorConfig exclude rules.
+			// Normalize the group to empty string for core-group resources ("v1" → "")
+			// to match the CollectorConfig apiGroups convention.
+			apiGroup := group
+			if len(groupVersion) == 1 {
+				apiGroup = ""
+			}
+			if tr.IsResourceExcluded(apiGroup, apiResource.Kind) {
+				klog.V(3).Infof("Skipping excluded resource [group: '%s' kind: %s]."+
+					" Matched CollectorConfig exclude rule.", apiGroup, apiResource.Kind)
```
smcavey: Tiniest of nits, then I'd say LGTM because this now prevents excluded informers from being initialized per @jlpadilla 's prior review.
`apiResource.Kind` -> `apiResource.Name` to match the kind log output in https://github.com/stolostron/search-collector/blob/main/pkg/informer/supportedResources.go#L141

Observed in my env:
```
I0702 17:37:46.930054       1 supportedResources.go:74] Allow resource [group: 'monitoring.coreos.com' kind: prometheusrules]. AllowList is empty.
I0702 17:37:46.930058       1 supportedResources.go:153] Skipping excluded resource [group: 'monitoring.coreos.com' kind: PrometheusRule]. Matched CollectorConfig exclude rule.
I0702 17:37:46.930061       1 supportedResources.go:74] Allow resource [group: 'monitoring.coreos.com' kind: podmonitors]. AllowList is empty.
I0702 17:37:46.930064       1 supportedResources.go:153] Skipping excluded resource [group: 'monitoring.coreos.com' kind: PodMonitor]. Matched CollectorConfig exclude rule.
I0702 17:37:46.930066       1 supportedResources.go:74] Allow resource [group: 'monitoring.coreos.com' kind: alertmanagers]. AllowList is empty.
I0702 17:37:46.930069       1 supportedResources.go:153] Skipping excluded resource [group: 'monitoring.coreos.com' kind: Alertmanager]. Matched CollectorConfig exclude rule.
I0702 17:37:46.930072       1 supportedResources.go:74] Allow resource [group: 'monitoring.coreos.com' kind: servicemonitors]. AllowList is empty.
I0702 17:37:46.930075       1 supportedResources.go:153] Skipping excluded resource [group: 'monitoring.coreos.com' kind: ServiceMonitor]. Matched CollectorConfig exclude rule.
I0702 17:37:46.930078       1 supportedResources.go:74] Allow resource [group: 'monitoring.coreos.com' kind: probes]. AllowList is empty.
I0702 17:37:46.930080       1 supportedResources.go:153] Skipping excluded resource [group: 'monitoring.coreos.com' kind: Probe]. Matched CollectorConfig exclude rule.
I0702 17:37:46.930083       1 supportedResources.go:74] Allow resource [group: 'monitoring.coreos.com' kind: prometheuses]. AllowList is empty.
I0702 17:37:46.930086       1 supportedResources.go:153] Skipping excluded resource [group: 'monitoring.coreos.com' kind: Prometheus]. Matched CollectorConfig exclude rule.
I0702 17:37:46.930088       1 supportedResources.go:74] Allow resource [group: 'monitoring.coreos.com' kind: thanosrulers]. AllowList is empty.
I0702 17:37:46.930091       1 supportedResources.go:74] Allow resource [group: 'monitoring.coreos.com' kind: alertmanagerconfigs]. AllowList is empty.
I0702 17:37:46.930094       1 supportedResources.go:153] Skipping excluded resource [group: 'monitoring.coreos.com' kind: AlertmanagerConfig]. Matched CollectorConfig exclude rule.
```

smcavey: Post change:
```
I0702 17:55:42.698991       1 supportedResources.go:74] Allow resource [group: 'monitoring.coreos.com' kind: alertmanagerconfigs]. AllowList is empty.
I0702 17:55:42.698995       1 supportedResources.go:153] Skipping excluded resource [group: 'monitoring.coreos.com' kind: alertmanagerconfigs]. Matched CollectorConfig exclude rule.
```

PR #909, opened 2026-07-02
ACM-21895: Refactor specific-kind transforms to use configurable collection

sahare:
## Summary

Refactors all 18 typed Builder functions and 5 unstructured-native Builder functions to call \`applyDefaultTransformConfig\` after their kind-specific property handling. This allows CollectorConfig rules (custom fields, \`collectConditions\`, \`collectAnnotations\`, \`collectAdditionalPrinterColumns\`) to take effect on these resource types, not just generic resources.

**Companion ticket:** ACM-21895

### Pattern (reference: PVC — already done before this PR)
```go
// BEFORE — CollectorConfig rules are ignored for this kind
func DeploymentResourceBuilder(d *v1.Deployment) *DeploymentResource {
    node := transformCommon(d)
    node.Properties["available"] = ...
    return &DeploymentResource{node: node}
}

// AFTER — CollectorConfig custom fields, conditions, annotations, printerColumns now work
func DeploymentResourceBuilder(d *v1.Deployment, r *unstructured.Unstructured) *DeploymentResource {
    node := transformCommon(d)
    node.Properties["available"] = ...
    node = applyDefaultTransformConfig(node, r)
    return &DeploymentResource{node: node}
}
```

### Typed builders refactored (18)
`deployment`, `daemonset`, `statefulset`, `replicaset`, `cronjob`, `persistentvolume`,
`argoapplication`, `application`, `channel`, `appdeployable`, `apphelmcr`,
`klusterletaddonconfig`, `placementbinding`, `placementrule`, `subscription`,
`policyreport`, `policy` (PolicyResourceBuilder), `deploymentconfig`

### Unstructured-native builders refactored (5 — no signature change)
`OperatorPolicyResourceBuilder`, `ConfigPolicyResourceBuilder`,
`CertPolicyResourceBuilder`, `KyvernoPolicyResourceBuilder`,
`VapBindingResourceBuilder`

### `transformer.go`
All 18 typed call sites updated to pass `event.Resource` (+ `event.AdditionalPrinterColumns` for CRD-backed kinds).

## Tests

Added `refactored_transforms_test.go` with **25 new tests** verifying that `applyDefaultTransformConfig` is correctly wired in for every refactored builder:
- Custom field extraction: all 18 typed + 3 unstructured-native builders
- `collectConditions`: Deployment (specific key) + `*.apps` wildcard
- `additionalColumns` pass-through: Subscription
- Early-return path: `CertPolicyResourceBuilder`
- Regression: Deployment-specific properties unaffected

[...]

PR #910, opened 2026-07-08
ACM-21895: Move specific-kind transform properties to genericResourceConfig (Phase 1)

sahare:
## Summary

Follow-up to #909. Moves hardcoded `node.Properties` assignments from specific-kind Builder functions into `defaultTransformConfig` in `genericResourceConfig.go`, and eliminates the Builder files and `transformer.go` cases for the fully-converted kinds.

**Jira:** ACM-21895

---

### Phase 1 — This PR: Full eliminations

Kinds where **all** properties are simple JSONPath extractions and `BuildEdges` is empty. The entire typed Builder is replaced by `GenericResourceBuilder` falling through from `transformer.go`.

| Kind | Properties moved | File deleted |
|---|---|---|
| Deployment (apps + extensions) | available, current, ready, desired | ✅ |
| DaemonSet (apps + extensions) | available, current, desired, ready, updated | ✅ |
| StatefulSet (apps) | current, desired | ✅ |
| ReplicaSet (apps + extensions) | current, desired | ✅ |
| DeploymentConfig (apps.openshift.io) | available, current, ready, desired | ✅ |
| PlacementRule (apps.open-cluster-management.io) | replicas | ✅ |
| Job (batch) | active, failed + adds: successful, completions, parallelism | ✅ |

**🐞 Also fixes a pre-existing bug in `job.go` where `parallelism` nil-check incorrectly tested `Spec.Completions` instead of `Spec.Parallelism`.**

### Phase 2 — Partial migrations (planned, not in this PR)

Kinds with edge logic that must stay in the Builder, but where simple property reads can move to config: Application, Channel, Deployable, ArgoApplication (destination/source/status fields), Policy (remediationAction, disabled, compliant), PVC (continue reference pattern), Subscription.

### Phase 3 — Framework extensions (planned, requires discussion)

To fully eliminate CronJob (`active = len(status.active)`), Subscription git annotation fallback, PlacementBinding formatted strings, AppHelmCR conditional sourceType switch. Would require additions like `DefaultValue`, `DataTypeArrayLength`, or multi-path conditional support on `ExtractProperty`.

---

## Test plan

```bash
DEPLOYED_IN_HUB=true go test ./... -failfast   # all packages pass
golangci-lint run ./...                          # 0 issues
```

Made with [Cursor](https://cursor.com)

## Summary

* **Refactor**
  * Switched multiple workload kind conversions to shared, generic extraction rules and removed several kind-specific transform implementations and dispatch paths.
* **New Features**
  * Added default-value support for JSONPath extractions, plus expanded default numeric extraction for common workload metrics (including Job-related fields).
* **Bug Fixes**
  * Ensured configurable collection merges keep the built-in base properties and only append custom properties.
* **Tests**
  * Updated and rewired transform tests to match the new property layout and numeric defaulting.
* **Documentation**
  * Updated the transforms README to clarify generic extraction vs per-kind logic.


pkg/transforms/genericResourceConfig.go
```diff
@@ -180,10 +180,90 @@ var defaultTransformConfig = map[string]ResourceConfig{
...
+	},
+	// StatefulSet — spec.replicas is *int32; absent when nil (same note as Deployment).
+	"StatefulSet.apps": {
 		properties: []ExtractProperty{
-			{Name: "active", JSONPath: `.status.active`},
-			{Name: "failed", JSONPath: `.status.failed`},
+			{Name: "current", JSONPath: `.status.replicas`, DataType: DataTypeNumber},
+			{Name: "desired", JSONPath: `.spec.replicas`, DataType: DataTypeNumber},
 		},
 	},
```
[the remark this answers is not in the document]
sahare: This deletion is intentional per jlpadilla's feedback on the previous PR https://github.com/stolostron/search-collector/pull/909#issuecomment-4897300161 requesting "cleanup of the specific transform files whenever possible." Kinds whose properties are fully expressed as JSONPath rules in genericResourceConfig.go and have no BuildEdges logic no longer need a dedicated file. We'll update the coding guideline to reflect this new pattern.

pkg/transforms/genericResourceConfig.go
```diff
@@ -180,10 +180,90 @@ var defaultTransformConfig = map[string]ResourceConfig{
 		},
 		extractAnnotations: true,
 	},
+	// Deployment and its extensions alias — replicas fields.
+	// NOTE: spec.replicas is a *int32 pointer; when nil (scale-to-zero or unset), the field is
+	// absent in the unstructured object and the property will not appear in search results.
+	// Previously the builder defaulted desired=0 when nil. Reviewers: is omitting preferred,
+	// or should we add a DefaultValue capability to ExtractProperty in a follow-up?
```
smcavey: Good question...In the case of deployment.apps, here's one instance of an impact: 

Image

smcavey: I deleted the desired key from acm-mcp-server data

jlpadilla: Defaulting to 0 would ensure this change doesn't introduce a breaking change.

@coderabbitai Can you help me check what we use in other places? Do we set a default or omit the property? I remember @smcavey made some changes about 6 months ago to maintain consistency.

smcavey: func applyDefaultTransformConfig has some specific DataTypeNumber handling in https://github.com/stolostron/search-collector/blob/main/pkg/transforms/common.go#L751-L875. But we might not even attempt to extract a num if len(result) on .spec.replicas is nil

smcavey: Can you rabbit eyes see that, @coderabbitai

smcavey: @jlpadilla The CRD documentation says it defaults to 1...
```
oc explain deployment.spec.replicas
GROUP:      apps
KIND:       Deployment
VERSION:    v1

FIELD: replicas <integer>

DESCRIPTION:
    Number of desired pods. This is a pointer to distinguish between explicit
    zero and not specified. Defaults to 1.
```

jlpadilla: @coderabbitai This repo is public, can you get common.go from here https://github.com/stolostron/search-collector/blob/main/pkg/transforms/common.go

jlpadilla: @smcavey Good point with the CRD doc, we should match that.

_The bunny went down a rabbit hole, but it was nice to have him here_

pkg/transforms/README.md
```diff
@@ -8,8 +8,10 @@ Here we document the properties collected for the kubernetes resources and the r
...
     - `kind (string), name (string), namespace (string), created (string), apigroup (string), apiversion (string), label ([]string)`
     - **Deprecated:** `selfLink`. It can be built from the properties above. We don't expect users to search for this.
-- Each transform file had a BuildNode() function where we define which properties we want to extract an index for the resource.
-- Our goal is to match the properties displayed from `oc get <resource> -o wide`, but we don't have a generic way to do this yet.
+- Properties are defined in one of two ways:
+  1. **`genericResourceConfig.go`** — for kinds whose properties are simple JSONPath extractions with no custom logic. Add a `"Kind.apiGroup"` entry to `defaultTransformConfig`. The `GenericResourceBuilder` and `applyDefaultTransformConfig` handle the rest. No per-kind file is needed.
+  2. **Per-kind file** (e.g. `pod.go`, `argoapplication.go`) — required when a kind needs `BuildEdges()` logic, computed properties (formatted strings, array lengths, conditional paths), or other handling that cannot be expressed as a JSONPath rule. The file implements `BuildNode()` and `BuildEdges()`.
```
jlpadilla, line 13: Going forward, option 1 is preferred.

pkg/transforms/transformer.go
```diff
@@ -294,14 +265,7 @@ func TransformRoutine(input chan *Event, output chan NodeEvent) {
...
 
-		case [2]string{"Job", "batch"}:
-			typedResource := batch.Job{}
-			err := runtime.DefaultUnstructuredConverter.
-				FromUnstructured(event.Resource.UnstructuredContent(), &typedResource)
-			if err != nil {
-				panic(err) // Will be caught by handleRoutineExit
-			}
-			trans = JobResourceBuilder(&typedResource, event.Resource)
+		// Job: all properties moved to genericResourceConfig.go; falls through to GenericResourceBuilder.
```
smcavey: I don't think we need all these comments, I think it's pretty inferred all the non-hardcoded kinds fall through to the default case of the switch

pkg/transforms/genericResourceConfig.go
```diff
@@ -180,10 +183,88 @@ var defaultTransformConfig = map[string]ResourceConfig{
...
+	},
+	"DaemonSet.apps": {
+		properties: []ExtractProperty{
+			{Name: "available", JSONPath: `.status.numberAvailable`, DataType: DataTypeNumber},
+			{Name: "current", JSONPath: `.status.currentNumberScheduled`, DataType: DataTypeNumber},
+			{Name: "desired", JSONPath: `.status.desiredNumberScheduled`, DataType: DataTypeNumber},
+			{Name: "ready", JSONPath: `.status.numberReady`, DataType: DataTypeNumber},
+			{Name: "updated", JSONPath: `.status.updatedNumberScheduled`, DataType: DataTypeNumber},
+		},
+	},
```
smcavey, line 212: PR looks good, we just need to make sure there's associated tests for each of these types and make sure that all these properties do get added to their typed nodes, e.g. https://github.com/stolostron/search-collector/blob/main/pkg/transforms/genericResource_test.go#L500-L514

pkg/transforms/genericResourceConfig.go
```diff
@@ -180,10 +183,88 @@ var defaultTransformConfig = map[string]ResourceConfig{
...
+	"DeploymentConfig.apps.openshift.io": {
+		properties: []ExtractProperty{
+			{Name: "available", JSONPath: `.status.availableReplicas`, DataType: DataTypeNumber},
+			{Name: "current", JSONPath: `.status.replicas`, DataType: DataTypeNumber},
+			{Name: "ready", JSONPath: `.status.readyReplicas`, DataType: DataTypeNumber},
+			{Name: "desired", JSONPath: `.spec.replicas`, DataType: DataTypeNumber},
+		},
+	},
+	"Job.batch": {
+		// active + failed were already here; adding successful, completions, parallelism.
```
jlpadilla, line 232: We can delete this line. No need to keep the history.

jlpadilla (approved):
/lgtm

PR #919, opened 2026-07-16
ACM-21892 - collectorconfig migration log warning

smcavey:
### Related Issue

https://issues.redhat.com/browse/ACM-21892

### Description of changes
- Inform users of existence of new collectorconfig and eventual deprecation of search-collector-config configmap

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated warnings for the `search-collector-config` configuration: the app now emits an explicit deprecation warning when the configuration is detected (not just when it’s missing), with the message tailored to the relevant feature setting.

### DIFF
--- a/pkg/informer/supportedResources.go
+++ b/pkg/informer/supportedResources.go
@@ -120,12 +120,21 @@ func SupportedResources(
 	kubeClient := config.GetKubeClient(config.GetKubeConfig())
 
 	// locate the search-collector-config ConfigMap
+	cmName := "search-collector-config"
 	cm, cmErr := kubeClient.CoreV1().ConfigMaps(config.Cfg.PodNamespace).
-		Get(ctx, "search-collector-config", machineryV1.GetOptions{})
+		Get(ctx, cmName, machineryV1.GetOptions{})
 	if cmErr != nil {
 		klog.Info("Collecting all resources. ConfigMap search-collector-config is not present.", cmErr)
 	} else {
-		klog.Warning("ConfigMap search-collector-config found and will be deprecated in the future by collectorconfigs.search.open-cluster-management.io. Consider migrating.")
+		if config.Cfg.FeatureConfigurableCollection {
+			klog.Warningf("ConfigMap %s found and feature Configurable Collection is enabled. "+
+				"%s and CollectorConfig may conflict with one another. The ConfigMap will be"+
+				"deprecated in a future release by collectorconfigs.search.open-cluster-management.io. "+
+				"You should migrate this configuration.", cmName, cmName)
+		} else {
+			klog.Warningf("ConfigMap %s found. It will be deprecated in the future by "+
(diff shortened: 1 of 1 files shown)


pkg/informer/supportedResources.go
```diff
@@ -124,6 +124,8 @@ func SupportedResources(
 		Get(ctx, "search-collector-config", machineryV1.GetOptions{})
 	if cmErr != nil {
 		klog.Info("Collecting all resources. ConfigMap search-collector-config is not present.", cmErr)
+	} else {
+		klog.Warning("ConfigMap search-collector-config found and will be deprecated in the future by collectorconfigs.search.open-cluster-management.io. Consider migrating.")
```
jlpadilla:
```suggestion
		klog.Warning("ConfigMap search-collector-config found. It will be deprecated in the future by collectorconfigs.search.open-cluster-management.io. Consider migrating.")
```

pkg/informer/supportedResources.go
```diff
@@ -124,6 +124,8 @@ func SupportedResources(
...
 	}
```
jlpadilla, line 138: I wonder if we should add a stronger warning if both ConfigMap and CollectorConfig are present.

smcavey: We could add an extra conditional if FEATURE_CONFIGURABLE_COLLECTION is enabled right there to warn their co-existence may interfere with one another

smcavey: If so, stronger warning, else the log we already have

jlpadilla: ok, let's use the feature flag, it's is easier.
It would be better to check for the CollectorConfig cr, but let's keep it simple and avoid the additional request.

smcavey:
```
		if config.Cfg.FeatureConfigurableCollection {
			klog.Warningf("ConfigMap %s found and feature Configurable Collection is enabled. "+
				"%s and Configurable Collection's collectorconfigs.search.open-cluster-management.io may " +
				"conflict with one another. Consider migrating.", cmName, cmName)
		} else {
			klog.Warningf("ConfigMap %s found. It will be deprecated in the future by "+
				"collectorconfigs.search.open-cluster-management.io. Consider migrating.", cmName)
		}
```
How's this verabage?

jlpadilla: Maybe stronger wording like this?
```go
klog.Warningf("ConfigMap %s found and feature Configurable Collection is enabled. "+
	"%s and CollectorConfig may conflict with one another. The ConfigMap will be" +
	"deprecated in a future release by collectorconfigs.search.open-cluster-management.io. "+
	"You should migrate this configuration.", cmName, cmName)
```

jlpadilla (approved):
/lgtm

PR #923, opened 2026-07-21
ACM-30184 - tls profile consistency

smcavey:
### Related Issue

https://issues.redhat.com/browse/ACM-30184

### Description of changes
- Added fetching and polling of tls profile config from ocm-tls-profile ConfigMap in open-cluster-management-agent-addon

## Summary by CodeRabbit

* **New Features**
  * Resolved TLS settings from the `ocm-tls-profile` ConfigMap (min TLS version and cipher-suite selection) with safe defaults and warnings for unrecognized entries.
  * Added TLS hot-reload by polling the ConfigMap and refreshing the sender’s HTTPS client when values change.
* **Tests**
  * Added unit tests for TLS version/cipher parsing, ConfigMap-to-TLS config behavior, and hot-reload signal handling.
* **Chores**
  * Updated Kubernetes/monitoring-related dependencies and aligned client-go version via the module replace directive.


pkg/send/httpsClient.go
```diff
@@ -25,48 +25,48 @@ import (
...
-		tr := &http.Transport{
-			TLSClientConfig: tlsCfg,
-		}
+	// Hub deployment:
+	// Load mounted certificates. If not found, use insecure TLS (development only).
+	caCert, err := os.ReadFile("./sslcert/tls.crt")
+	cert, err2 := tls.LoadX509KeyPair("./sslcert/tls.crt", "./sslcert/tls.key")
+	if err != nil || err2 != nil {
+		klog.Error("WARNING: Using insecure TLS connection. Couldn't load certs ", err, err2)
+		tlsCfg.InsecureSkipVerify = true
```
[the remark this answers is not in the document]
smcavey: Thanks for raising this but it's a preexisting issue, maybe should be fixed, but not in the scope of the Jira

pkg/config/tlsConfig.go
```diff
@@ -0,0 +1,120 @@
...
+	configv1 "github.com/openshift/api/config/v1"
+	openshifttls "github.com/openshift/controller-runtime-common/pkg/tls"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
+)
+
+const (
+	tlsProfileConfigMap = "ocm-tls-profile"
+	tlsProfileNamespace = "open-cluster-management-agent-addon"
```
jlpadilla: This is the default, but it could change. It's better to use $POD_NAMESPACE which gets updated according to where the collector is deployed.

pkg/config/tlsConfig.go
```diff
@@ -0,0 +1,120 @@
...
+)
+
+// GetTLSConfig reads the ocm-tls-profile ConfigMap and returns a *tls.Config.
+// Falls back to the OpenShift Intermediate TLS profile if the ConfigMap can't be read.
+func GetTLSConfig() *tls.Config {
+	cm, err := GetKubeClient(GetKubeConfig()).CoreV1().ConfigMaps(tlsProfileNamespace).
+		Get(context.TODO(), tlsProfileConfigMap, metav1.GetOptions{})
+	if err != nil {
+		klog.Warningf("Could not read ConfigMap %s/%s, falling back to Intermediate profile: %v",
```
jlpadilla:
```suggestion
		klog.Warningf("Could not read ConfigMap %s/%s, falling back to Intermediate TLS profile: %v",
```

pkg/config/tlsConfig.go
smcavey: Hmm okay, I went with this approach because on the hub this namespace and configmap also exist. So our hub and managed clusters could fetch the tls profile from the same place in the same manner. But if that namespace isn't static then that might introduce inconsistency in where the hub and managed clusters access this configmap, in which case we might as well just query the apiserver directly like the operator's approach at this

pkg/config/tlsConfig.go
```diff
@@ -0,0 +1,120 @@
...
+)
+
+const (
+	tlsProfileConfigMap = "ocm-tls-profile"
+	tlsProfileNamespace = "open-cluster-management-agent-addon"
+)
+
+// GetTLSConfig reads the ocm-tls-profile ConfigMap and returns a *tls.Config.
+// Falls back to the OpenShift Intermediate TLS profile if the ConfigMap can't be read.
+func GetTLSConfig() *tls.Config {
```
jlpadilla, line 26: The collector is deployed on Hub and Managed clusters. Looks like the current implementation only covers the managed clusters. For the Hub, I expect to see a path where TLS profile can be loaded from ENV variable, similar to indexer and API projects.

smcavey: https://github.com/stolostron/search-collector/pull/923/changes#diff-8b6f6c7d36e07cf85f26054a62d2f344ef5b7cac20d5479f40669e3cbfc7cf87R29 is where we first fetch this config regardless of hub vs. managed

pkg/config/tlsConfig.go
jlpadilla: The Hub only has this namespace when hubSelfManagement (local-cluster) is enabled.

smcavey:
per discussion with @jlpadilla pivoting to get hub cluster tls profile config via env vars and operator. managed cluster config obtained via the configmap but by using the PodNamespace config instead of hard-coded namespace

jlpadilla (approved):
/lgtm

PR #936, opened 2026-08-03
ACM-30184 - metrics endpoint tls profile consistency

smcavey:
### Related Issue

https://issues.redhat.com/browse/ACM-30184

### Description of changes
- Added ...

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved server startup by applying the configured TLS settings for HTTPS connections.
  * The server now automatically uses HTTPS when a certificate is available and falls back to HTTP otherwise.
  * Added clearer startup and error messages for both HTTP and HTTPS modes.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -3,7 +3,6 @@
 package server
 
 import (
-	"k8s.io/klog/v2"
 	"net/http"
 	"os"
 	"os/signal"
@@ -14,6 +13,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/stolostron/search-collector/pkg/config"
 	"github.com/stolostron/search-collector/pkg/metrics"
+	"k8s.io/klog/v2"
 )
 
 func StartAndListen() {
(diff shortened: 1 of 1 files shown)


jlpadilla:
🤖 I'm an AI agent

## Review: ACM-30184 — metrics endpoint TLS profile consistency

The approach is solid — applying the cluster TLS profile to the metrics server is the right thing to do. I found one must-fix issue and two should-fix items before this is safe to merge.

---

### 🔴 Must-fix: Operator probes and ServiceMonitor will break on the hub-local collector

The operator configures three things to talk to port 5010 over plain HTTP, all in `search-v2-operator`:

1. **Liveness probe** — `Scheme: "HTTP"` (`create_collectordeployment.go:58`)
2. **Readiness probe** — `Scheme: "HTTP"` (`create_collectordeployment.go:47`)
3. **CollectorServiceMonitor** — explicitly overrides to `Scheme: "http"` and sets `TLSConfig: nil` (`create_metrics_monitor.go:71-72`)

The hub-local collector has `/sslcert` mounted (the operator does it in `create_collectordeployment.go:37`), so after this PR the server will start HTTPS. All three of the above will get TLS errors — the kubelet will kill the pod on failed health checks and Prometheus won't be able to scrape metrics.

**This needs a companion operator PR** updating the probe schemes to `HTTPS` and fixing `CollectorServiceMonitor` to use `https` + TLS config before (or coordinated with) this merging. Managed-cluster collectors are unaffected since they don't have `/sslcert` mounted.

---

### 🟡 Should-fix: `os.Stat` doesn't fail closed (unresolved CodeRabbit comment)

The current else-branch catches **all** `os.Stat` errors including permission errors, silently starting HTTP when the cert exists but can't be read:

```go
// current
} else {
    klog.Warning("TLS certificates not found, starting HTTP server")
    ...
}

// suggested
} else if os.IsNotExist(err) {
    klog.Warning("TLS certificates not found, starting HTTP server")
    ...
} else {
    klog.Fatal("Cannot inspect TLS certificate: ", err)
}
```

---

### 🟡 Should-fix: `ReadTimeout` not set

`ReadHeaderTimeout` limits only header reads. For a metrics endpoint this is low-risk but worth fixing as a one-liner:

```go
ReadTimeout:       time.Duration(config.Cfg.HTTPTimeout) * time.Millisecond,
ReadHeaderTimeout: time.Duration(config.Cfg.HTTPTimeout) * time.Millisecond,
```

---

### Minor nits (non-blocking)
- PR description still has the "Added ..." placeholder — worth filling in.
- The two commits could be squashed; the first set `TLSConfig` without cert loading so it was incomplete on its own.

/hold — waiting on companion operator PR for probe + ServiceMonitor updates.

PR #954, opened 2026-09-10
Add userCount for OpenShift Groups

ruici-h:
User Story: https://redhat.atlassian.net/browse/ACM-42679
Adds `userCount` for `Group.user.openshift.io` resources by counting entries in `.users`.
- Returns `0` for empty, missing, or null `users`
- Adds unit coverage for the new behavior

Tested with `make test` and verified against a 5.0 test hub.

## Summary by CodeRabbit

- **New Features**
  - Added support for deriving array lengths with the `sliceLen` data type.
  - OpenShift Group resources now expose `userCount` based on the number of users.
  - Missing, null, or empty user lists are handled appropriately, including configured defaults where applicable.
- **Bug Fixes**
  - Non-array values are no longer stored as slice-length properties and generate a warning.

### DIFF
--- a/pkg/transforms/common.go
+++ b/pkg/transforms/common.go
@@ -857,6 +857,17 @@ func applyDefaultTransformConfig(node Node, r *unstructured.Unstructured, additi
 						prop.Name, kind, group, r.GetName(), val,
 					)
 				}
+			} else if prop.DataType == DataTypeCount {
+				if arr, ok := val.([]interface{}); ok {
+					node.Properties[prop.Name] = int64(len(arr))
+				} else if val == nil {
+					node.Properties[prop.Name] = int64(0)
+				} else {
+					klog.V(1).Infof(
+						"Unable to count prop [%s] from [%s.%s] Name: [%s], not an array: %T",
+						prop.Name, kind, group, r.GetName(), val,
+					)
+				}
 			} else if prop.DataType == DataTypeBoolean {
 				// DataTypeBoolean: convert to string representation
 				if b, ok := val.(bool); ok {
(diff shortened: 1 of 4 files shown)


smcavey:
I only with jsonPath had a built in len func or something so we didn't have to add a new datatype, but the implementation is solid. I just want to stew on it a bit

pkg/transforms/common.go
```diff
@@ -857,6 +857,17 @@ func applyDefaultTransformConfig(node Node, r *unstructured.Unstructured, additi
 						prop.Name, kind, group, r.GetName(), val,
 					)
 				}
+			} else if prop.DataType == DataTypeCount {
+				if arr, ok := val.([]interface{}); ok {
+					node.Properties[prop.Name] = int64(len(arr))
+				} else if val == nil {
+					node.Properties[prop.Name] = int64(0)
```
jlpadilla: Here we should honor the default value set in the configuration.
```suggestion
					node.Properties[prop.Name] = prop.DefaultValue
```

pkg/transforms/genericResourceConfig.go
```diff
@@ -34,6 +34,7 @@ const (
 	DataTypeNumber    DataType = "number"
 	DataTypeMapString DataType = "mapString"
 	DataTypeBoolean   DataType = "boolean"
+	DataTypeCount     DataType = "count"
```
jlpadilla, line 37: Wondering if `length` or `arrayLen` is a better name.

smcavey: `sliceLen` or `length` have my vote. not `array` because we've already established use of `slice`

ruici-h: `sliceLen` makes sense since this specifically operates on slices. I’ll update it.

smcavey:
/lgtm
you'll just have to --signoff your commits

smcavey:
Looks like it's mostly just the last else condition with the log that's uncovered, and that's not really worth testing a log 
image
so I think we just override sonar

PR #956, opened 2026-09-15
ACM-42679: index Group users for Search filtering

ruici-h:
### Story Issue
https://redhat.atlassian.net/browse/ACM-42679
This completes request 1 of the story. Request 2 (`userCount`) was added in #954.

### Description of changes
Added a `users` property to the `Group.user.openshift.io` transform config, indexing Group member usernames as a Search array so Console can perform reverse lookups for the User details → Groups view.

To avoid duplicating the membership array across managed clusters, `users` is collected only by the hub collector. `userCount` continues to be collected on all clusters.

Added `TestGroup_Users` coverage for:
- multiple members
- a single member
- empty users
- absent users
- null users
- hub collector includes users
- managed collector omits users while preserving userCount
- The existing `TestGroup_UserCount` coverage remains unchanged.
### Related PRs
- search-indexer: adds the GIN index for data -> 'users'
https://github.com/stolostron/search-indexer/pull/744
- search-v2-api: uses users for filtering while omitting the membership array from normal OpenShift Group search results
https://github.com/stolostron/search-v2-api/pull/786


smcavey:
I'd like to do some testing of the scale proposed by the Jira. If there's 25k groups, there's got to be at least 25k users. 25k database records each storing some subset of 25k users could be a lot of storage. I'd like to test if we can store users and groups as separate resources (to remove the duplication) and handle the membership via edges. Then we can not only answer the question which groups does Bob belong in, but also which users are in group X.

ruici-h:
> I'd like to do some testing of the scale proposed by the Jira. If there's 25k groups, there's got to be at least 25k users. 25k database records each storing some subset of 25k users could be a lot of storage. I'd like to test if we can store users and groups as separate resources (to remove the duplication) and handle the membership via edges. Then we can not only answer the question which groups does Bob belong in, but also which users are in group X.

Followed up in Slack — we discussed this further and compared the array + GIN and edge approaches at scale.

Based on the testing and real customer data, array + GIN had a smaller storage footprint with comparable query performance, so we’re planning to continue with that approach.

I’ll continue with the remaining work - adding the GIN index in `search-indexer` and updating API-side change so the full users[] array isn’t returned on every Group search result.

pkg/transforms/genericResourceConfig.go
```diff
@@ -236,6 +238,8 @@ var defaultTransformConfig = map[string]ResourceConfig{
 	},
 	"Group.user.openshift.io": {
 		properties: []ExtractProperty{
+			// Group membership is only collected from the hub.
+			{Name: "users", JSONPath: `.users[*]`, DataType: DataTypeSlice, hubOnly: true},
```
jlpadilla: Maybe an alternative to reduce load is to not include this as part of the default configuration here. Instead, console could add it to the integration CollectorConfig when needed.

@ruici-h This implementation is correct 👍 , I'm just thinking of alternatives to reduce the load.

